PR #1, opened 2019-10-16
Fetch access token via OAuth2

p-janik:
Scope of this ticket is to allow to fetch an access token via OAuth2 (PKCE).


README.md
```diff
@@ -1,2 +1,48 @@
...
+const client = new Client({
+  clientId: "YOUR_CLIENT_ID",
+  redirectUri: "YOUR_REDIRECT_URI",
+  scopes: ["transactions"]
+});
+
+client.getAuthUri().then(uri => {
+  /* redirect your user to this uri so he can provide his credentials */
+});
+
```
johannespfeiffer: Of course this is just an example - but we only should do this if the user is _not_ already logged in.

lib/types.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export declare class Client {
+  private oauth2Client: ClientOAuth2;
+  private state: string;
+
+  constructor(opts: ClientOpts);
+
+  public getAuthUri(opts: GetAuthUriOpts): Promise<string>;
+  public getToken(callbackUri: string, opts: GetTokenOpts): Promise<ClientOAuth2.Token>;
+}
```
johannespfeiffer: We do not need this. Just run `tsc` and look into `dist/client.d.ts` which will create sth. similar already.

package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "kontist-sdk",
+  "version": "0.0.1",
+  "description": "Kontist client SDK",
+  "main": "dist/index.js",
+  "types": "lib/types.ts",
```
johannespfeiffer: should be `dist/index.d.ts`

lib/types.ts
p-janik: true, corrected with https://github.com/kontist/sdk/pull/1/commits/4f75a79b29463f3c788741cfe8fe8157d4dcfa84

PR #2, opened 2019-10-17
Extend client with raw graphQL queries

p-janik:


README.md
```diff
@@ -53,3 +53,23 @@ app.get(CALLBACK_PATH, (req, res) => {
...
+```typescript
+const query = `{
+  viewer {
+    mainAccount {
+      id
+    }
+  }
+}`;
+
+const result = client.rawQuery(query);
```
hihuz: should it be `await client.rawQuery` ? 🤔

lib/client.ts
```diff
@@ -65,10 +71,53 @@ export class Client {
...
+      token = this.oauth2Client.createToken(accessToken, refreshToken, data);
+    } else {
+      token = this.oauth2Client.createToken(accessToken, data);
+    }
+
+    this.token = token;
+    return token;
+  };
+
+  public rawQuery = async (
```
hihuz: I was wondering if it would be nice to have a separate helper for mutation 🤔but I'm not sure if it would be easy to do with this `graphql-request` module, I don't know it.

It's just a thought for the future anyway, for now we can do with just this one helper it will be totally fine

hihuz:
Ah, I thought you were only going to implement the base raw query in this PR, then I'll wait with my review until you are done with the rest 🙂

johannespfeiffer:
Had to do a few minor adjustments to make it testable in browser environments.

lib/client.ts
```diff
@@ -17,58 +24,156 @@ export class Client {
...
+    const result: Query = await this.rawQuery(query);
+
+    const transactions = (
+      (result &&
+        result.viewer &&
+        result.viewer.mainAccount &&
+        result.viewer.mainAccount.transactions &&
+        result.viewer.mainAccount.transactions.edges) ||
+      []
+    ).map(node => node.node);
```
johannespfeiffer: Looks a bit ugly, can we e.g. use lodash until we can use TypeScript 3.7 and the nullish coalescing operator `??`? so it might look like `_.get(result, "viewer.mainAccount.transactions.edges")`

lib/client.ts
```diff
@@ -17,58 +24,156 @@ export class Client {
...
+      `Bearer ${this.token.accessToken}`
     );
-    return tokenData;
+
+    const { data } = await this.graphQLClient.rawRequest(query, variables);
+
+    return data;
+  };
+
+  public getAllTransactions = async (): Promise<Array<TransactionListItem>> => {
```
johannespfeiffer: What about moving this to a separate file into a method `getAll()` or `fetch()` and then include it here in the client? The user could call it with `await client.transactions.getAll()`

lib/client.ts
p-janik: sure, done in https://github.com/kontist/sdk/pull/2/commits/e9013b343956d1cafd39dfcf0df4ecad6a580603

lib/client.ts
p-janik: sure, i had a plan to split it, this file cannot grow indefinitely :)

PR #3, opened 2019-10-18
Separate auth and graphql clients

p-janik:


lib/graphql/model.ts
```diff
@@ -0,0 +1,13 @@
...
+export class Model {
+  protected client: GraphQLClient;
+
+  constructor(graphqlClient: GraphQLClient) {
+    this.client = graphqlClient;
+  }
+
+  async fetchAll(): Promise<Array<Object>> {
+    throw new Error("Must be overridden by subclasses!");
+  }
```
johannespfeiffer: Can we make the class and this method `abstract`?

PR #5, opened 2019-10-20
Setup verifier during init

johannespfeiffer:
Setup `verifier` (optional) during the initialization and save it in the `Client` object. This simplifies the calls of `getAuthUri` and `getToken` a bit. Let me know what you think.


p-janik:
@johannespfeiffer I think it has perfect sense, thanks for this improvement

PR #9, opened 2019-10-25
Add pagination for transaction resource

p-janik:
Todo:
- [x] update kontist.dev sdk page reagrding changes in this PR

### DIFF
--- a/lib/graphql/model.ts
+++ b/lib/graphql/model.ts
@@ -3,14 +3,12 @@ import { ResultPage } from "./resultPage";
 import { FetchOptions } from "./types";
 import { IFetch } from "./interfaces";
 
-export class Model<T> implements IFetch<T> {
+export abstract class Model<T> implements IFetch<T> {
   protected client: GraphQLClient;
 
   constructor(graphqlClient: GraphQLClient) {
     this.client = graphqlClient;
   }
 
-  async fetch(args?: FetchOptions): Promise<ResultPage<T>> {
-    throw new Error("Must be overridden by subclasses!");
-  }
+  abstract async fetch(args?: FetchOptions): Promise<ResultPage<T>>;
 }
--- a/lib/graphql/model.ts
+++ b/lib/graphql/model.ts
@@ -1,13 +1,16 @@
 import { GraphQLClient } from "./client";
+import { ResultPage } from "./resultPage";
+import { FetchOptions } from "./types";
+import { IFetch } from "./interfaces";
 
-export class Model {
+export class Model implements IFetch {
   protected client: GraphQLClient;
 
   constructor(graphqlClient: GraphQLClient) {
     this.client = graphqlClient;
   }
 
-  async fetchAll(): Promise<Array<Object>> {
+  async fetchAll(args?: FetchOptions): Promise<ResultPage> {
     throw new Error("Must be overridden by subclasses!");
   }
 }
--- a/lib/graphql/model.ts
+++ b/lib/graphql/model.ts
@@ -10,7 +10,7 @@ export class Model implements IFetch {
     this.client = graphqlClient;
   }
 
-  async fetchAll(args?: FetchOptions): Promise<ResultPage> {
+  async fetch(args?: FetchOptions): Promise<ResultPage> {
     throw new Error("Must be overridden by subclasses!");
   }
 }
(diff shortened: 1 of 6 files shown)


lib/graphql/resultPage.ts
```diff
@@ -0,0 +1,20 @@
+import { PageInfo } from "./schema";
+import { IFetch } from "./interfaces";
+
+export class ResultPage {
+  items: Array<Object>;
+  nextPage?: () => Promise<ResultPage>;
+  previousPage?: () => Promise<ResultPage>;
+
+  constructor(model: IFetch, items: Array<Object>, pageInfo: PageInfo) {
+    this.items = items;
```
johannespfeiffer: Any chance we can use generics here? 
```
export class ResultPage<T> {
  items: Array<T>;
  nextPage?: () => Promise<ResultPage<T>>;
  previousPage?: () => Promise<ResultPage<T>>;

  constructor(model: IFetch, items: Array<T>, pageInfo: PageInfo) {
```

then later in `transactions.ts` sth. like

```
export class Transaction extends Model {
   async fetch(args?: FetchOptions): Promise<ResultPage<Transaction>> {
```

lib/graphql/transaction.ts
```diff
@@ -5,50 +5,58 @@ import {
   TransactionsConnectionEdge
 } from "./schema";
 import { Model } from "./model";
+import { FetchOptions } from "./types";
+import { ResultPage } from "./resultPage";
+
+const safeGet = (source: any, path: any, defaultValue: any) =>
+  get(source, path, defaultValue) || defaultValue;
```
johannespfeiffer: Just out of curiosity: When do we need that?

lib/graphql/transaction.ts
```diff
@@ -5,50 +5,58 @@ import {
...
-                  mandateNumber
-                  paymentMethod
-                  category
-                  userSelectedBookingDate
-                  purpose
-                  bookingType
-                  invoiceNumber
-                  invoicePreviewUrl
-                  invoiceDownloadUrl
-                  documentType
```
johannespfeiffer: Did we loose them? Should be `documentNumber`, `documentPreviewUrl`, `documentDownloadUrl`, ...

lib/graphql/types.ts
```diff
@@ -0,0 +1,6 @@
+export type FetchOptions = {
+  first?: number | null;
+  last?: number | null;
+  before?: String | null;
+  after?: String | null;
```
johannespfeiffer: Why do we allow `number` and `null` as a type? The `?` already allows the type to be `undefined`.

johannespfeiffer: I think we can use just `string` instead of `String`.

lib/graphql/transaction.ts
```diff
@@ -5,50 +5,58 @@ import {
...
-  async fetchAll(): Promise<Array<TransactionEntry>> {
-    const query = `{
-      viewer {
+  async fetch(args?: FetchOptions): Promise<ResultPage> {
+    const query = `
+      query fetchTransactions ($first: Int, $last: Int, $after: String, $before: String) {
+        viewer {
           mainAccount {
-            transactions {
+            transactions(first: $first, last:$last, after: $after, before: $before) {
```
johannespfeiffer: nit: missing space in `last:$last`

lib/graphql/model.ts
```diff
@@ -1,13 +1,16 @@
...
   protected client: GraphQLClient;
 
   constructor(graphqlClient: GraphQLClient) {
     this.client = graphqlClient;
   }
 
-  async fetchAll(): Promise<Array<Object>> {
+  async fetch(args?: FetchOptions): Promise<ResultPage> {
     throw new Error("Must be overridden by subclasses!");
   }
```
johannespfeiffer: Can we make the class and this method abstract instead?

lib/graphql/transaction.ts
p-janik: `get` returns `defaultValue` only when property from patch is undefined. Does not work like this in other falsy cases

lib/graphql/transaction.ts
p-janik: true, it was renamed

lib/graphql/types.ts
p-janik: blindly followed `before` and `after`

p-janik:
Thanks, created https://github.com/kontist/kontist.github.io/pull/12 to update the docs

PR #10, opened 2019-10-25
Create shortcut methods for (batch)transfer creation/confirmation

p-janik:


README.md
```diff
@@ -88,6 +88,29 @@ while (transactionPage.nextPage) {
...
+  note: <optional_description>,
+  e2eId: <optional_e2eId>,
+});
+```
+
+To confirm a transfer:
+
+```typescript
+const result = await client.models.transfer.confirm(
+  <transfer_id>,
```
johannespfeiffer: This has been recently renamed to `confirmation_id`.

lib/graphql/transfer.ts
```diff
@@ -0,0 +1,51 @@
...
+          e2eId: "${transfer.e2eId}", 
+          note: "${transfer.note}" 
+        }
+      ) {
+        id
+        status
+        amount
+        note
+        iban
+        e2eId
```
johannespfeiffer: Does only return a confirmation_id now.

README.md
p-janik: I don't see it on latest backend

p-janik: corrected in https://github.com/kontist/sdk/pull/10/commits/982f621b4a7286b7b394f9f6e5e6cfb653f14261

lib/graphql/transfer.ts
p-janik: shall we make it consistent also on the backend side?

lib/graphql/transfer.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  async confirmMany(
+    confirmationId: string,
+    authorizationToken: string
+  ): Promise<BatchTransfer> {
+    const query = `mutation {
+      confirmTransfers(
+        confirmationId: "${confirmationId}"
+        authorizationToken: "${authorizationToken}"
+      ) {
```
dmytroyarmak: I think it would be more secure to pass `variables` alongside `query` instead of interpolating values it JS (the same that you do for `createMany`). It should prevent possible injections.

lib/graphql/transfer.ts
dmytroyarmak: Unfortunately it's not yet merged to `develop`. It was done in scope of `feature/5743-mutations-fo-all-transfer-types` branch that Lera is working on.

PR #13, opened 2019-10-31
Add OAuth2 authorization_code & refresh_token to SDK

p-janik:
Looks like https://kontist.dev/sdk/#login-backend-application is up to date

### DIFF
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -8,6 +8,13 @@ export class Auth {
   private _token: ClientOAuth2.Token | null = null;
   private verifier?: string;
 
+  /**
+   * Client OAuth2 module instance.
+   *
+   * @param baseUrl  Kontist API base url
+   * @param opts     OAuth2 client data including at least clientId, redirectUri,
+   *                 scopes, state and clientSecret or verifier (for PKCE).
+   */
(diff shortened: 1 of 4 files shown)


lib/auth.ts
```diff
@@ -111,6 +133,11 @@ export class Auth {
     return token;
   };
```
johannespfeiffer: Not directly scope of this ticket, but do you think we can simplify this e.g. with sth. like this?
```
return this._token = this.oauth2Client.createToken(..._.compact(accessToken, refreshToken, tokenType), {});
```

johannespfeiffer: You are right. Seems there are a couple of open issues for this problem:
https://github.com/microsoft/TypeScript/issues/22061
https://github.com/microsoft/TypeScript/issues/22112
https://github.com/microsoft/TypeScript/issues/5453

Proposed workaround might look like this, but I would not use it here, seems it brings more confusion than it is worth.
```typescript
const method = this.oauth2Client.createToken;
this.oauth2Client.createToken(
  ..._.compact([accessToken, refreshToken, tokenType, {}]) 
  as Parameters<typeof method>
);```

PR #14, opened 2019-10-31
Add user and account models

p-janik:


lib/graphql/index.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Transaction } from "./transaction";
+import { GraphQLClient } from "./client";
+
+export type Models = {
+  transaction: Transaction;
+  transfer: Transfer;
+  account: Account;
+  user: User;
+};
+
```
johannespfeiffer: Can we just use type interference here and remove this type?
Bildschirmfoto 2019-11-01 um 13 19 04

then in `client.ts` either use `  public models: ReturnType<typeof getModels>;` or even just via constructor assignment like this (which btw could simplify unit testing as well) :

```typescript
export class Client {
  constructor(
    opts: ClientOpts,
    baseUrl = opts.baseUrl || KONTIST_API_BASE_URL,
    public auth = new Auth(baseUrl, opts),
    public graphQL = new GraphQLClient(`${baseUrl}/api/graphql`, auth),
    public models = getModels(graphQL)
    ) { }
}
```

lib/graphql/queries.ts
```diff
@@ -0,0 +1,133 @@
+export const GET_USER = `query {
```
johannespfeiffer: I like the idea to outsource all graphql queries, but can we discuss if it makes sense to keep them (as constants) in their corresponding files instead of one giant `queries.ts`?

lib/graphql/user.ts
```diff
@@ -0,0 +1,20 @@
+import { User as UserModel, Query } from "./schema";
+import { Model } from "./model";
+import { ResultPage } from "./resultPage";
+import { GET_USER } from "./queries";
+
+export class User extends Model<UserModel> {
+  async fetch(): Promise<ResultPage<UserModel>> {
+    throw new Error("You are allowed only to fetch your details.");
```
johannespfeiffer: Very nice, in the future this might be possible (for certain clients) so it is good to keep the same method names.

lib/graphql/queries.ts
p-janik: you're right, I'll move it back them back to corresponding files

PR #16, opened 2019-10-31
Add auth methods for MFA

hihuz:
In this PR we are adding a public method to the `Auth` class which will initiate the MFA flow by creating a challenge, and then wait for the challenge to be confirmed / denied / expired before resolving / rejecting.

If you want to test the whole flow, you will need to have setup a device with push notifications enabled for your user (that means real device or android simulator), and a proper setup for oauth clients in your local db.

You can test in browsers with the [API Console](https://github.com/kontist/kontist.github.io/tree/master/console-src), just add: 
```
try {
  const confirmedToken = await kontistClient.auth.getMFAConfirmedToken();
  console.log({ confirmedToken });
} catch (err) {
  console.log({ err });
}
```
in `AuthenticateUser.tsx` after the call to `kontistClient.auth.fetchToken`, and also make sure that you have the compiled `@kontist/client` lib with the changes from this branch in your node_modules

To test in node environment, just copy paste the node example from the readme of this repo in a `server.js` file, add the same try catch block as above, and run it with `node server.js`


README.md
```diff
@@ -194,3 +194,34 @@ const result = await client.models.transfer.confirmMany(
...
+MFA is available once you have installed the Kontist application and paired your device in it.
+
+The following steps are necessary to complete the MFA procedure:
+1. initiate the procedure by creating a challenge (Kontist SDK exposes a method to do that)
+2. click the push notification you received on your phone, it will open the Kontist application
+3. login (if applicable) and confirm the MFA by clicking on the corresponding button
+
+Kontist SDK exposes a method to initiate the MFA flow after you successfully received the initial access token:
+
+```typescript
```
hihuz: I wasn't sure if I should have included these examples in the regular `Usage` paragraphs, or as a standalone here 🤔

lib/auth.ts
```diff
@@ -1,12 +1,26 @@
...
 import { btoa } from "abab";
-import { ClientOpts, GetTokenOpts, GetAuthUriOpts } from "./types";
+import {
+  ClientOpts,
+  GetAuthUriOpts,
+  Challenge,
+  ChallengeStatus,
+  HttpMethod
+} from "./types";
+import "cross-fetch/polyfill";
```
hihuz: This is a dependency of `graphql-request` that we already use. I think it's best to not include another lib and stick to that one, any opinion?

lib/auth.ts
```diff
@@ -111,6 +126,81 @@ export class Auth {
     return token;
   };
 
+  /**
+   * Perform a request against Kontist REST API
+   */
+  private request = async (path: string, method: HttpMethod, body?: string) => {
```
hihuz: I was contemplating extracting this to another class / module, what do you think? I also couldn't think of a proper name for this extracted class while implementing so I left it here for now.

lib/auth.ts
```diff
@@ -111,6 +126,81 @@ export class Auth {
...
+    resolve: Function,
+    reject: Function
+  ) => async () => {
+    const challenge = await this.request(
+      `${MFA_CHALLENGE_PATH}/${pendingChallenge.id}`,
+      HttpMethod.GET
+    );
+
+    if (new Date(challenge.expiresAt) < new Date()) {
+      if (this.challengePollIntervalId) {
```
hihuz: So timeout ids in node have type `NodeJS.Timeout` and in browser they have `number`, so I defined this property like so: `private challengePollIntervalId?: ReturnType<typeof setInterval>;` 

Which works fine but I couldn't find a way to initialise it, if I do `private challengePollIntervalId: ReturnType<typeof setInterval>; = 1234` compiler yells as type `number` is not compatible with type `NodeJS.Timeout`.

So here I have to do these ugly checks 😬any fluent typescripter would have an idea how I can avoid this?

lib/auth.ts
```diff
@@ -111,6 +126,81 @@ export class Auth {
...
+
+      this.setToken(confirmedToken);
+      resolve(confirmedToken);
+    }
+  };
+
+  /**
+   * Create an MFA challenge and request a confirmed access token when verified
+   */
+  public createMFAChallenge = async () => {
```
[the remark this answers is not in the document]
johannespfeiffer: Maybe sth. with `enhanceTokenByMFA` or `upgradeTokenByMFA`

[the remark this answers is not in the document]
hihuz: I do like `enhanceTokenByMFA` 🙂I'll wait until tomorrow to see if there are other ideas

tonyskn:
Looks good when it comes to supporting the 2FA UI flow we already implemented in Webapp 👍 

We would call `client.auth.createMFAChallenge()`, transition our local UI state to "pending confirmation", then move forward depending on what this Promise settles on...

tonyskn:
One general suggestion though:

Would you consider throwing custom `Error` classes instead of a vanilla `Error` with a plain text message?

Something like `throw new MfaChallengeExpiredError()` (or at least `throw new Error(errorsCodes.MFA_CHALLENGE_EXPIRED)`) would make it easier for clients to dispatch to specific error handlers without pattern matching over a `string` message...

lib/auth.ts
p-janik: I agree

lib/auth.ts
```diff
@@ -133,6 +147,81 @@ export class Auth {
...
+   * Perform a request against Kontist REST API
+   */
+  private request = async (path: string, method: HttpMethod, body?: string) => {
+    if (!this.token) {
+      throw new Error("User unauthorized");
+    }
+
+    const requestUrl = new URL(path, this.baseUrl).href;
+
+    const response = await fetch(requestUrl, {
```
p-janik: should we throw an error when `response.ok === false`?

lib/auth.ts
```diff
@@ -133,6 +147,81 @@ export class Auth {
...
+    } else if (challenge.status === ChallengeStatus.DENIED) {
+      if (this.challengePollIntervalId) {
+        clearInterval(this.challengePollIntervalId);
+      }
+      reject(new Error("Challenge denied"));
+    } else if (challenge.status === ChallengeStatus.VERIFIED) {
+      if (this.challengePollIntervalId) {
+        clearInterval(this.challengePollIntervalId);
+      }
+      const { token: confirmedToken } = await this.request(
```
johannespfeiffer: Are we are returning the plain token string here? Because `fetchToken` returns a `ClientOAuth2.Token` which e.g. has methods to refresh etc.
Does it make sense to align the output type?

README.md
johannespfeiffer: We should add that to https://kontist.dev/sdk as well.

hihuz: Yeah I was planning to do it but I wanted first to get some feedback on it here so that I don't have to make changes in both places 🙂

lib/auth.ts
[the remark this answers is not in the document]
hihuz: @tonyskn since you might be actually using this sdk helper in the webapp at some point, do you have any opinion on naming for the method to call? 🙂

lib/auth.ts
p-janik: you need to have those checks cause clearInterval accepts non-falsy value, right? 
You could also cast it to NodeJS.Timeout like 
```typescript
clearInterval(this.challengePollIntervalId as NodeJS.Timeout);
```
which should be fine for the browser also cause this type won't be in the final bundle file.

Maybe you can also extract it to small function so it won't look that ugly anymore :)

lib/auth.ts
```diff
@@ -133,6 +147,81 @@ export class Auth {
...
+    const challenge = await this.request(
+      `${MFA_CHALLENGE_PATH}/${pendingChallenge.id}`,
+      HttpMethod.GET
+    );
+
+    if (new Date(challenge.expiresAt) < new Date()) {
+      if (this.challengePollIntervalId) {
+        clearInterval(this.challengePollIntervalId);
+      }
+      reject(new Error("Challenge expired"));
```
p-janik: can you add explicit `return` to those reject/resolve calls? 🙏🏻

lib/auth.ts
p-janik: I would say it's fine to keep it here for now

lib/auth.ts
hihuz: > you need to have those checks cause clearInterval accepts non-falsy value, right?

Yep, exactly

> You could also cast it to NodeJS.Timeout like ...

This is a great idea and it works 🙂 I think I will extract a type like this:
`type IntervalID = ReturnType<typeof setInterval>;`
and use it for the property definition `private challengePollIntervalId?: IntervalID;` and cast my timeouts to it `clearInterval(this.challengePollIntervalId as IntervalID);`

Thank you for the suggestion 🙏

lib/auth.ts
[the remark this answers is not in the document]
tonyskn: `createMFAChallenge()` actually returns a confirmed token, not a challenge, so this name is a bit misleading :)

No strong opinions about what to rename it to, though...
What about `getMFAConfirmedToken` ?

lib/auth.ts
hihuz: Have changed it in this commit: ce5eeca110fffd68790f3e6d601736f7b4011a79

lib/auth.ts
hihuz: Sure 🙂 added here: e9ecb7de0e6e2fe3a4187f87602dcbc8ef575e9f

lib/auth.ts
hihuz: Absolutely right, I overlooked that, I have changed it here: 0184496a204b7b5b76369436cbfb94bf61a1c8be

lib/auth.ts
hihuz: Great idea, I have added this check here: ed834c70d4864e03a6d94fe79925dd8193815249

lib/auth.ts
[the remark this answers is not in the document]
hihuz: Seems good to me, I went with `getMFAConfirmedToken` in the end: 78816e03b8574e8d001531085a7aa6b15aa5dd0a

hihuz:
> Would you consider throwing custom Error classes instead of a vanilla Error with a plain text message?

Great suggestion 🙏

So I have setup new custom errors in this commit: 755a314d999f9ae762a2b790a7192dfdfb98b67b

 I'm not sure if it corresponds to what you had in mind, feel free to give feedback.

I have created a base `KontistSDKError` and several custom errors extending it.

When they are thrown you will get the following in the browser:
Screenshot 2019-11-01 at 18 29 54
* `type` which you can specify and will default to `KONTIST_SDK_ERROR`, it's an enum
* `status` which is of type `number` you can optionally specify (for non network-related errors it doesn't really make sense)
* `name` which will be the name of the class throwing

So you can base your error handling logic on any of the above fields. you also have the regular `error.message` which is now an enum also.


Regarding `name` property, I wanted to override it in the base class using `this.name = this.constructor.name` but apparently typescript prevents it, see: https://github.com/Microsoft/TypeScript/issues/14099

So the best option I found was to do:
```
    Object.setPrototypeOf(this, ChallengeDeniedError.prototype);
    this.name = this.constructor.name;
```
in the constructor of every child class. Not very convenient, but it will work for now. If better ideas I will definitely update it.

lib/errors.ts
```diff
@@ -0,0 +1,73 @@
+type ErrorOpts = {
+  message?: string;
+  status?: number;
+  type?: ErrorType;
+};
+
+export enum ErrorType {
```
johannespfeiffer: Do we really need this `ErrorType` enum?
I think we might just use `instanceOf` or do I miss sth.?

```typescript
type ErrorOpts = {
    message?: string;
    status?: number;
};

class KontistSDKError extends Error {
    public status?: number;

    constructor(opts: ErrorOpts = {}) {
        super(opts.message);
        this.status = opts.status;
    }
}

class ChallengeExpiredError extends KontistSDKError {
    constructor(opts: ErrorOpts = {}) {
        super(opts);
        Object.setPrototypeOf(this, ChallengeExpiredError.prototype);
    }
}

const e = new ChallengeExpiredError({ status: 42 });

console.log(e instanceof ChallengeExpiredError ?
    'is instanceof ChallengeExpiredError' : 'not instanceof ChallengeExpiredError')

throw e;
```

Outputs:
```
is instanceof ChallengeExpiredError
/Users/johannes/tmp_error/index.js:37
throw e;
^

ChallengeExpiredError
    at ChallengeExpiredError.KontistSDKError [as constructor] (/Users/johannes/tmp_error/index.js:19:28)
    at new ChallengeExpiredError (/Users/johannes/tmp_error/index.js:29:28)
    at Object.<anonymous> (/Users/johannes/tmp_error/index.js:35:9)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1025:10)
    at internal/main/run_main_module.js:17:11 {
  status: 42
```

hihuz: No you're right, we don't really need it, I just wasn't sure what would be the preferred way of handling these errors so I added a bunch of different possible ways. Let's stick to `instanceof` and remove this `type` property.

a9c9f2419fc2d39b751a7219658eeffb86920359

lib/auth.ts
```diff
@@ -133,6 +152,82 @@ export class Auth {
...
+    resolve: Function,
+    reject: Function
+  ) => async () => {
+    const challenge = await this.request(
+      `${MFA_CHALLENGE_PATH}/${pendingChallenge.id}`,
+      HttpMethod.GET
+    );
+
+    if (new Date(challenge.expiresAt) < new Date()) {
+      clearInterval(this.challengePollIntervalId as IntervalID);
```
dmytroyarmak: this line is repeated multiple times, would it be useful to extract it to a separate method?

lib/auth.ts
```diff
@@ -133,6 +152,82 @@ export class Auth {
...
+   * Create an MFA challenge and request a confirmed access token when verified
+   */
+  public getMFAConfirmedToken = async () => {
+    const challenge = await this.request(MFA_CHALLENGE_PATH, HttpMethod.POST);
+
+    return new Promise((resolve, reject) => {
+      this.challengePollIntervalId = setInterval(
+        this.pollChallengeStatus(challenge, resolve, reject),
+        this.challengePollInterval
+      );
```
dmytroyarmak: This polling implementation can have multiple parallel requests if user has very slow internet connection or `challengePollInterval` is less than request time. Is it ok for us?

lib/auth.ts
```diff
@@ -133,6 +152,82 @@ export class Auth {
...
+
+      const token = this.setToken(confirmedToken);
+      return resolve(token);
+    }
+  };
+
+  /**
+   * Create an MFA challenge and request a confirmed access token when verified
+   */
+  public getMFAConfirmedToken = async () => {
```
dmytroyarmak: Do we need to allow user to stop polling after some time? Does it make sense to provide method for it or accept `timeout`?

tonyskn: Actually, yes:
In the webapp, we need a way to cancel polling if User decides to abandon the "please confirm your account" screen and go back to the Login form.

lib/auth.ts
```diff
@@ -196,14 +196,15 @@ export class Auth {
       HttpMethod.GET
     );
```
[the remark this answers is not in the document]
hihuz: Good catch, you are totally right of course 🙂I think I like rejecting a little bit more in this case because otherwise the user could be stuck while multiple requests are failing without being notified of it. Rejecting allows a clean "try again later" approach. What do you think?

I have changed here: bcb4212999739c49099d17771c0946e760b653aa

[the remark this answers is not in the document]
dmytroyarmak: I think the problem with internet is very rare on desktop, so it would be fine to just reject because most likely something is wrong with the server 👍

PR #19, opened 2019-11-06
Add silent auth token refresh for browsers

hihuz:
In this PR we are adding an auth method to refresh the user's access token using `response_mode=web_message` implemented in our backend recently: https://github.com/kontist/figure-backend/pull/5821

You can test the flow by first authenticating using regular login, and then calling the new `refreshTokenSilently()` method:
```
    const url = await kontistClient.auth.getAuthUri();
    window.location.href = url;
    ...
    const token = await kontistClient.auth.refreshTokenSilently();
```

I have shamelessly forked @dmytroyarmak PKCE test application and added these methods if you want to get an easier time testing: https://github.com/hihuz/pkce-demo
just make sure that data in `config.js` is correct against your database: you need an oauth client with proper redirect uri, scopes (`users`) and grant_types (`authorization_code`)

### DIFF
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -174,7 +174,9 @@ export class Auth {
   /**
    * Refresh auth token silently for browser environments
    */
-  public refreshTokenSilently = async (timeout?: number): Promise<ClientOAuth2.Token> => {
+  public refreshTokenSilently = async (
+    timeout?: number
+  ): Promise<ClientOAuth2.Token> => {
     if (!document || !window) {
       throw new SilentAuthorizationError({
         message:
@@ -191,7 +193,9 @@ export class Auth {
 
     try {
       const code = await authorizeSilently(iframeUri, this.baseUrl, timeout);
-      const fetchTokenUri = `${document.location.origin}?code=${code}&state=${this.state}`;
+      const fetchTokenUri = `${
+        document.location.origin
+      }?code=${code}&state=${encodeURIComponent(this.state)}`;
       const token = await this.fetchToken(fetchTokenUri);
 
       return token;
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -174,7 +174,7 @@ export class Auth {
   /**
    * Refresh auth token silently for browser environments
    */
-  public refreshTokenSilently = async (): Promise<ClientOAuth2.Token> => {
+  public refreshTokenSilently = async (timeout?: number): Promise<ClientOAuth2.Token> => {
     if (!document || !window) {
       throw new SilentAuthorizationError({
         message:
@@ -190,7 +190,7 @@ export class Auth {
     });
 
     try {
-      const code = await authorizeSilently(iframeUri, this.baseUrl);
+      const code = await authorizeSilently(iframeUri, this.baseUrl, timeout);
       const fetchTokenUri = `${document.location.origin}?code=${code}&state=${this.state}`;
       const token = await this.fetchToken(fetchTokenUri);
 
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -21,7 +21,7 @@ import {
   ChallengeDeniedError,
   MFAConfirmationCanceledError,
   UserUnauthorizedError,
-  SilentAuthorizationError,
+  RenewTokenError,
   KontistSDKError
 } from "./errors";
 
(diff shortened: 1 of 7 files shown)


lib/utils/authorizeSilently.ts
```diff
@@ -0,0 +1,52 @@
...
+export const authorizeSilently = (uri: string, origin: string) => {
+  return new Promise((resolve, reject) => {
+    let timeoutId: TimeoutID;
+    const iframe = document.createElement("iframe");
+    iframe.style.display = "none";
+    document.body.appendChild(iframe);
+
+    const cleanup = () => {
+      window.removeEventListener("message", onMessageHandler);
+      if (iframe && iframe.parentNode) {
```
hihuz: Screenshot 2019-11-05 at 15 51 16

🤷‍♂

lib/utils/authorizeSilently.ts
```diff
@@ -0,0 +1,52 @@
...
+        return reject(event.data.error);
+      }
+    };
+
+    window.addEventListener("message", onMessageHandler);
+    if (iframe) {
+      iframe.src = uri;
+    }
+
+    timeoutId = setTimeout(() => {
```
hihuz: does it seem useful to timeout after some time if server doesn't send message? is 10 secs appropriate?

lib/utils/authorizeSilently.ts
p-janik: `document.body.removeChild(iframe);`?

hihuz: I was a bit reluctant to do that because if the SDK is used in a context where other `iframe` are attached to the body by any other code it could be problematic I think 🤔

lib/auth.ts
```diff
@@ -120,6 +131,37 @@ export class Auth {
...
+    const iframeUri = await this.getAuthUri({
+      query: {
+        prompt: "none",
+        response_mode: "web_message"
+      }
+    });
+
+    try {
+      const code = await authorizeSilently(iframeUri, this.baseUrl);
+      const fetchTokenUri = `${document.location.origin}?code=${code}&state=${this.state}`;
```
dmytroyarmak: Because state can have arbitrarily value it would be safer to encode it by using `URLSearchParams` or `encodeURIComponent`. Otherwise if it has `=`,`&` or `#` it will brake the url.

lib/utils/authorizeSilently.ts
```diff
@@ -0,0 +1,52 @@
...
+        return;
+      }
+
+      cleanup();
+
+      if (!event.data.error) {
+        const { code } = event.data.response;
+        return resolve(code);
+      } else {
+        return reject(event.data.error);
```
dmytroyarmak: What if `event.data` is `undefined`?

lib/utils/authorizeSilently.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+    };
+
+    window.addEventListener("message", onMessageHandler);
+    if (iframe) {
```
dmytroyarmak: Can `iframe` become `null` or `undefined` here?

lib/utils/authorizeSilently.ts
dmytroyarmak: Yes, it's totally make sense. 
Should we make it configurable to be sure that it will be appropriate for app that uses it?

hihuz: Yeah I agree, good idea 🙂 I have added a parameter for this to the method: 3ca8ece9625143e2009d047a254beeaa4b81c4fc

lib/auth.ts
hihuz: Oh wow great suggestion, I double checked manually and it was indeed breaking 🙏 This commit should take care of it: 879418e491e76666b836958526344f737b8f3937

lib/utils/authorizeSilently.ts
hihuz: Hmm I thought it wasn't necessary to check for it since it is defined in our backend 🤔But I guess it doesn't hurt to double check at SDK's end. What about this commit 3eddfacaf534d923be2e50b502684f6c5636a1cf is it fine like so?

lib/utils/authorizeSilently.ts
hihuz: Not as far as I can tell, I added this check because at some point I had the same error as in my comment above `Object is possibly 'null'` but not anymore apparently 🤔 so I have removed the check 👍

lib/utils/authorizeSilently.ts
dmytroyarmak: it looks great 🙏

hihuz:
@johannespfeiffer That is a good idea 👍 

`token.refresh` is defined by our third party `client-oauth2`, so I think I'd rather stick to `client.auth.refresh()`

What do you think about this implementation? 11bc0919a27f9bd6a709c212f8300b3bbedfc92e

There are quite a few of changes there and I'm not sure I like all of the naming, but the most important part is the public method that we expose to users and document so for that I think `client.auth.refresh` is fine

PR #24, opened 2019-11-08
GraphQL error handling

hihuz:
This PR addresses: https://github.com/kontist/figure-backend/issues/5854

Our GraphQL backend now appends error details when an API error is thrown in the `extensions` field (See GQL spec details about it here: https://graphql.github.io/graphql-spec/June2018/#sec-Errors).

It is a good opportunity to standardise the way we handle them in the SDK.

Here is how they are thrown after this PR:
Screenshot 2019-11-07 at 17 38 43

### DIFF
--- a/lib/utils/serializeGraphQLError.ts
+++ b/lib/utils/serializeGraphQLError.ts
@@ -1,0 +1,36 @@
+import { ErrorOpts } from "../errors";
+
+type GraphQLAPIError = {
+  response?: {
+    errors?: [
+      {
+        message?: string;
+        extensions?: {
+          status?: number;
+          type?: string;
+        };
+      }
+    ];
+  };
+};
+
+export const serializeGraphQLError = (graphQLError: GraphQLAPIError): ErrorOpts => {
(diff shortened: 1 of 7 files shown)


lib/utils/serializeGraphQLError.ts
```diff
@@ -0,0 +1,36 @@
...
+        };
+      }
+    ];
+  };
+};
+
+export const serializeGraphQLError = (graphQLError: GraphQLAPIError): ErrorOpts => {
+  const errorOptions: ErrorOpts = {};
+
+  const errorDetails =
```
hihuz: if https://github.com/kontist/js-sdk/pull/22 gets merged before this I can simplify this quite a bit 🙂

johannespfeiffer: done

hihuz: Thank you 🙏 I changed it here: badbc826c2493dedc1958e1833a47a95af89cb69

PR #26, opened 2019-11-08
Add a changelog

johannespfeiffer:
Please have a look if that format makes sense, based on https://keepachangelog.com/en/1.0.0/.
On release we would just move the items from `[Unreleased]` into sth like `## [0.0.18] 2019-11-08`


johannespfeiffer:
> Maybe a small improvement that we can have would be to have actual links to the PRs?

Good point, added the links.

PR #28, opened 2019-11-11
Breakdown auth module

hihuz:
Addresses: https://github.com/kontist/figure-backend/issues/5867

Our auth module was becoming bloated, specifically regarding the 2 multi-factor auth procedures (push notification verification, and device binding).
The idea was to have a leaner base `Auth` class which would delegate:
* push notification MFA logic to a `PushNotification` class instance available under `client.auth.push`
* device binding logic to a `DeviceBinding` class instance available under `client.auth.device`

Additional considerations:
* Both MFA methods (`push` and `device`) need to make requests against Kontist REST API so I have extracted the `request` helper to its own module/class
* Both MFA methods need to share existing token management logic and state so I have extracted that to its own `TokenManager` class, with instance available under `client.auth.tokenManager` and injected to both MFA class constructors
* Most token / auth related logic has been moved to the more specific `TokenManager`, so if we consider for instance the `fetchToken` method, it was previously available under `client.auth.fetchToken` and now it will be `client.auth.tokenManager.fetchToken` introducing a breaking change. In the current implementation, I have added simple aliases in the `Auth` class for these like so:
```
  public fetchToken = async (
    callbackUri: string
  ): Promise<ClientOAuth2.Token> => this.tokenManager.fetchToken(callbackUri);
```
This would ensure that we do not break anything, but I'd like some feedback regarding how to approach that, do you think that we should:
* introduce the breaking changes now without any aliases in the `Auth` class (this will force SDK users to do some changes, even if they are fairly straightforward)
* add the aliases as mentioned above, indicating that they are `@deprecated`
* add the aliases above without mentioning anything (pretty much this would mean that `TokenManager` remains an implementation detail for us and we do not encourage using it directly

I wouldn't worry about any breaking changes to push notification / device binding as I don't believe anybody is using that yet.

Will update documentation over at https://github.com/kontist/kontist.github.io once we are happy with all naming and separation of modules.


lib/auth/auth.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.device = new DeviceBinding(this.tokenManager, request);
+    this.push = new PushNotificationMFA(this.tokenManager, request);
+  }
+
+  /**
+   * @deprecated use tokenManager method directly instead
+   *
+   * Build a uri to which the user must be redirected for login.
+   */
+  public getAuthUri = async (opts: GetAuthUriOpts = {}): Promise<string> =>
```
hihuz: Please let me know what you think is the best way to proceed regarding these breaking changes (taking into consideration ease of use for SDK users and maintainability for us), should we:
* add aliases to the `Auth` class like in the current implementation, indicating that they are `@deprecated` (this would prevent anything from breaking, but would discourage usage of the new methods)
* add aliases to the `Auth` class like in the current implementation, **without** indicating that they are deprecated, making the `TokenManager` class more of an implementation detail (downside being that we have to maintain a mapping between `TokenManager` methods and `Auth` aliases)
* do not add aliases here and introduce the breaking changes directly, forcing users to change their code to `client.auth.tokenManager.xxxx` (upside being that there aren't too many users of the SDK currently)

hihuz: As per a discussion with @johannespfeiffer, for now I will keep the `@deprecated` approach adding some warnings in dev mode and specifying when the methods will be removed (thanks for the suggestions 🙏) :
Screenshot 2019-11-12 at 10 34 45


Still open for discussion / other ideas of course.

hihuz:
Great, flipping to `!== "production"` makes total sense, and extracting the tests as well (I was planning to do it once we settled on the deprecation approach, but then forgot, so thanks a lot 🙂)

PR #29, opened 2019-11-11
Update changelog with released 0.18.0 changes

hihuz:


hihuz:
I thought it was on purpose but I don't think it's a big deal anyway, we can consider that we are on minor 0.18 now 🙂

PR #30, opened 2019-11-12
Export errors and IResultPageInterface

tonyskn:
Not sure if we should export `IResultPageInterface` directly or expose a `types` module...


tonyskn:
Sounds good 👍 
Will merge and tag a `0.19.0` on master

PR #32, opened 2019-11-12
Add authorizeSilently unit tests

hihuz:
In the initial PR (https://github.com/kontist/js-sdk/pull/19) to add PKCE silent token refresh, I did not add specific unit tests for the `authorizeSilently` function.

I think it's valuable to add some tests there (added benefit is to also increase our test coverage at the same time)

Before:
Screenshot 2019-11-12 at 19 09 26
Screenshot 2019-11-12 at 19 10 04

After:
Screenshot 2019-11-12 at 19 08 14
Screenshot 2019-11-12 at 19 09 57


lib/utils/authorizeSilently.ts
```diff
@@ -29,7 +29,7 @@ export const authorizeSilently = (
 
       cleanup();
 
-      if (event.data && !event.data.error) {
```
hihuz: this wasn't really a valid / useful check before, so writing these tests had at least the value of making me notice it and fix it 🙂

lib/utils/authorizeSilently.ts
```diff
@@ -29,7 +29,7 @@ export const authorizeSilently = (
...
+      if (event.data && event.data.response) {
```
johannespfeiffer: nit: could this now be as well look like this?
```typescript
      if (event.data?.response) {
        const { code } = event.data.response;
        return resolve(code);
      } else {
        const error = event.data?.error ??
          new RenewTokenError({
            message: "Invalid message received from server"
          });
        return reject(error);
      }
```

Same for `if (iframe && iframe.parentNode)` etc.

hihuz: True, still not used to go for these optional chainings as my go to 🙂

hihuz:
Great points, and thanks for adding the test case 🙂 

> Can we remove the check for timeoutId? Seems clearTimeout works with undefined as well - can you confirm that?

I can confirm that it would work with undefined but I believe I added this check to appease typescript compiler, I'll double check and remove it if possible

> Can we add a small testcase when the iframe is missing?

Good idea, yeah 🙂

lib/utils/authorizeSilently.ts
hihuz: Changed here: 7e5c5451b494a2be8af8a355ed7a8459c0c560cd 👍

johannespfeiffer:
Great idea with `iframe.remove()` even better than an additional testcase 👍

hihuz:
I have made the following changes:

* removed check for timeoutId existence, it doesn't bother the compiler: 1b5145eef68c0cba5ab7ca11ff831230d29f659f
* used optional chaining where suggested: 7e5c5451b494a2be8af8a355ed7a8459c0c560cd
* I started implementing the test cases for the missing branches, and after fighting for some time I realised we could get rid of this branching logic by using simply `iframe.remove()` instead 🙂 a6452b39c937056436a43b16bb72321bc3d26731

So we now have 100% for this file:
Screenshot 2019-11-13 at 13 48 10

PR #33, opened 2019-11-13
Add transaction subscription logic

hihuz:
Addresses: https://github.com/kontist/figure-backend/issues/5892

In this PR we are adding a `client.models.transaction.subscribe()` method which uses gql subscriptions to notify SDK users when new transactions are received (see the readme changes for explanations about the usage)

Notes:
* In the end we went with an "observable" style approach instead of "event listeners", it seemed a bit more reasonable during the implementation
* we are attempting to reconnect automatically on behalf of the user when the websocket connection is closed

### DIFF
--- a/lib/graphql/client.ts
+++ b/lib/graphql/client.ts
@@ -4,7 +4,7 @@ import { Variables } from "graphql-request/dist/src/types";
 import * as ws from "ws";
 
 import { Auth } from "../auth";
-import { RawQueryResponse } from "./types";
+import { RawQueryResponse, SubscriptionType, Unsubscribe } from "./types";
 import { serializeGraphQLError } from "../utils";
 import { GraphQLError, UserUnauthorizedError } from "../errors";
 import { GraphQLClientOpts } from "../types";
@@ -12,29 +12,27 @@ import { GraphQLClientOpts } from "../types";
 const WebSocket = typeof window === "undefined" ? ws : window.WebSocket;
 
 type Subscriptions = {
-  [key: string]: Function;
+  [key: number]: {
+    id: number;
+    query: string;
+    type: SubscriptionType;
+    handler: Function;
+    unsubscribe: Unsubscribe;
+  };
 };
 
 export class GraphQLClient {
   private auth: Auth;
   private client: GQLClient;
-  private subscriptionClient: SubscriptionClient;
+  private subscriptionEndpoint: string;
+  private subscriptionClient?: SubscriptionClient | null;
   private subscriptions: Subscriptions = {};
+  private subscriptionId: number = 0;
 
   constructor({ auth, endpoint, subscriptionEndpoint }: GraphQLClientOpts) {
     this.auth = auth;
     this.client = new GQLClient(endpoint);
-    this.subscriptionClient = new SubscriptionClient(
-      subscriptionEndpoint,
-      {
-        lazy: true,
-        reconnect: true,
-        connectionParams: () => ({
-          Authorization: `Bearer ${this.auth.tokenManager.token?.accessToken}`
-        })
-      },
-      WebSocket
-    );
+    this.subscriptionEndpoint = subscriptionEndpoint;
   }
 
   /**
@@ -62,9 +60,56 @@ export class GraphQLClient {
   };
 
   /**
-   * Subscribe to a topic and call the handler when new data is received
+   * Create a subscription client
    */
-  public subscribe = (query: string, topic: string, handler: Function) => {
+  private createSubscriptionClient = (): SubscriptionClient => {
+    if (!this.auth.tokenManager.token) {
+      throw new UserUnauthorizedError();
+    }
+
+    return new SubscriptionClient(
+      this.subscriptionEndpoint,
+      {
+        lazy: true,
+        connectionParams: () => ({
+          Authorization: `Bearer ${this.auth.tokenManager.token?.accessToken}`
+        })
+      },
+      WebSocket
+    );
+  };
+
+  /**
+   * Handle disconnection by refreshing access token and resubscribing
+   * all previously active subscriptions
+   */
+  private handleDisconnection = async (): Promise<void> => {
+    await this.auth.tokenManager.refresh();
+    Object.values(this.subscriptions).forEach(subscription => {
+      this.subscribe(
+        subscription.query,
+        subscription.type,
+        subscription.handler,
+        subscription.id
+      );
+    });
+  };
+
+  /**
+   * Subscribe to a topic and call the handler when new data or an error is received
+   */
+  public subscribe = (
+    query: string,
+    type: SubscriptionType,
+    handler: Function,
+    subscriptionId?: number
+  ): Unsubscribe => {
+    if (!this.subscriptionClient) {
+      this.subscriptionClient = this.createSubscriptionClient();
+
+      this.subscriptionClient.onDisconnected(this.handleDisconnection);
+    }
+
     const subscription = this.subscriptionClient.request({
       query
     });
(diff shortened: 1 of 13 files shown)


lib/graphql/client.ts
```diff
@@ -1,19 +1,42 @@
...
+import * as ws from "ws";
 
 import { Auth } from "../auth";
-import { RawQueryResponse } from "./types";
+import { RawQueryResponse, SubscriptionType, Unsubscribe } from "./types";
 import { serializeGraphQLError } from "../utils";
 import { GraphQLError, UserUnauthorizedError } from "../errors";
+import { GraphQLClientOpts } from "../types";
+
+type Subscriptions = {
```
hihuz: these types are kinda local so I thought it was fine to keep them in this file, but I can definitely extract them if it looks more proper

lib/graphql/client.ts
```diff
@@ -37,4 +60,109 @@ export class GraphQLClient {
...
+  };
+
+  /**
+   * Handle disconnection:
+   *
+   * 1. Refresh access token
+   * 2. Destroy existing subscription client
+   * 3. Resubscribe all previously active subscriptions
+   */
+  private handleDisconnection = async (): Promise<void> => {
```
hihuz: Unfortunately, the `subscriptions-transport-ws` doesn't provide any helper or logic to manage reconnection, so this forced me to keep track of all the active subscriptions at all times so that we can "resubscribe" after a disconnection.

It adds a bit of overhead but it's not too bad so I still went ahead and did it.

lib/graphql/transaction.ts
```diff
@@ -55,9 +70,22 @@ export class Transaction extends IterableModel<TransactionEntry> {
   async fetch(args?: FetchOptions): Promise<ResultPage<TransactionEntry>> {
```
hihuz: The changes to this method are unrelated to my feature, I use prettier to format for me and it did these formatting changes, I can remove them if necessary

lib/graphql/transaction.ts
```diff
@@ -55,9 +70,22 @@ export class Transaction extends IterableModel<TransactionEntry> {
...
 
-    const pageInfo = result?.viewer?.mainAccount?.transactions?.pageInfo ?? { hasNextPage: false, hasPreviousPage: false};
+    const pageInfo = result?.viewer?.mainAccount?.transactions?.pageInfo ?? {
+      hasNextPage: false,
+      hasPreviousPage: false
+    };
     return new ResultPage(this, transactions, pageInfo);
   }
+
+  subscribe(handler: (event: SubscriptionEvent) => any): Unsubscribe {
```
hihuz: So this is what we will officially expose and document I think.

it has a "observable" style to it in that we subscribe passing a handler and it returns the method to unsubscribe, hope it looks fine? from a user's perspective it seemed the most simple approach to me

tests/graphql/client.spec.ts
```diff
@@ -1,84 +1,563 @@
...
+  });
+
+  describe("when access token is missing", () => {
+    before(() => {
+      client.auth.tokenManager._token = undefined;
+    })
+
+    it("should still return proper connection params", () => {
+      expect(client.graphQL.getConnectionParams()).to.deep.equal({
+        Authorization: "Bearer undefined"
```
hihuz: This test is very silly, this can't happen in practice but I needed to add this case for 100% coverage on the feature 🤷‍♂

lib/graphql/client.ts
```diff
@@ -37,4 +60,109 @@ export class GraphQLClient {
...
+  public subscribe = (
+    query: string,
+    type: SubscriptionType,
+    handler: Function,
+    subscriptionId?: number
+  ): Unsubscribe => {
+    if (!this.subscriptionClient) {
+      this.subscriptionClient = this.createSubscriptionClient();
+
+      this.subscriptionClient.onDisconnected(this.handleDisconnection);
```
johannespfeiffer: This can lead to a `UnhandledPromiseRejectionWarning` as `handleDisconnection` is async and its rejections (e.g. error from `await this.auth.tokenManager.refresh();`) are not catched anywhere. To reproduce one can just stop the backend while a subscription is active.

lib/graphql/client.ts
johannespfeiffer: Agree.

lib/graphql/client.ts
```diff
@@ -37,4 +60,109 @@ export class GraphQLClient {
       throw new GraphQLError(serializeGraphQLError(error));
     }
   };
+
+  /**
+   * Return subscription client connection params
+   */
+  private getConnectionParams = (): ConnectionParams => ({
+    Authorization: `Bearer ${this.auth.tokenManager.token?.accessToken}`
+  });
```
johannespfeiffer: Inlining this could even simplify the code a bit, let us discuss that if you want.

lib/graphql/client.ts
johannespfeiffer: Makes sense, but if the token refresh fails we should throw an error which the user can handle then.

tests/graphql/client.spec.ts
johannespfeiffer: Or let's skip that helper method at all (see my inline comment).

hihuz:
Great comments thank you very much 🙂I will address these first: https://github.com/kontist/js-sdk/pull/36

And then come back to this PR right after 👍

lib/graphql/client.ts
hihuz: You are absolutely correct, it was even inlined at first, and I extracted it for helping with other tests or something similar at some point but some other refactorings must have made this extraction pointless.

Great catch 👍
Changed in this commit: 12b9b2e12b288e8c1c0bc88da28e5c03bc422e75

lib/graphql/client.ts
```diff
@@ -37,4 +60,109 @@ export class GraphQLClient {
...
+      unsubscribe
+    };
+
+    return this.unsubscribe(id);
+  };
+
+  /**
+   * Unsubscribe to a topic
+   */
+  public unsubscribe = (subscriptionId: number): (() => void) => () => {
```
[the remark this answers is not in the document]
hihuz: Agreed, I went with `createUnsubscriber`, if it makes sense. Changed in this commit: 6d35089fc78b4263a76e3c12d4b3eecd1cda9a64

lib/graphql/types.ts
```diff
@@ -7,4 +7,18 @@ export type FetchOptions = {
...
+  newTransaction = "newTransaction"
+}
+
+export type SubscriptionEvent = {
+  data?: any;
+  error?: Error;
+}
+export type TransactionSubscriptionEvent = SubscriptionEvent & {
+  data?: Transaction;
+};
```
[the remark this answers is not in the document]
hihuz: So in the end, with the new implementation, I don't think these types are relevant anymore, I just return the new entity plainly, so in the transaction model it goes like this:
```
  subscribe(
    onNext: (event: TransactionEntry) => any,
    onError?: (error: Error) => any
  ): Subscription {
  ...
}
```
The change is in this commit: 990c62c08f270767a53e00b0434f935100669e45

hihuz:
> The refresh can fail - would be nice if we can pass this error down to the sdk user.

True, great catch regarding the fact that the error could not be caught with the previous implementation.
In my opinion the best way to handle this is to call the subscribed `onError` handlers when it occurs, would you agree? here is how it would look like: 
67de95da193ac9c4b86712c10132d47da7da22a6

> Great idea with using Observables. Can we try to match this interface even a bit more?

Absolutely, I also went back and forth between one handler and 2. I think we can keep the `onError` optional, what do you think?
I also completely agree with you regarding `onComplete` it doesn't make much sense for a subscription as it is not supposed to be completed at any point

I have updated the implementation and tests in this commit: 990c62c08f270767a53e00b0434f935100669e45

lib/graphql/client.ts
hihuz: Great point, see my below comment regarding that, changes are in this commit: 67de95da193ac9c4b86712c10132d47da7da22a6

lib/graphql/client.ts
hihuz: Great point, see my below comment regarding that, changes are in this commit: 67de95da193ac9c4b86712c10132d47da7da22a6

hihuz:
> During the tests I was missing the baseSubscriptionUrl option. As discussed it almost never makes sense to have a custom baseUrl but use baseSubscriptionUrl from production. No good idea here yet.

So this is the last point I haven't addressed yet. I have thought of 2 alternatives to the current implementation but none of them are fully satisfying:

1. have `baseUrl` parameter accept an URL without protocol (e.g. `api.kontist.com`), and define both URLs based on that, along with a `secure: boolean` option to go `wss` + `https` instead of `ws` + `http`

2. just replace `http` by `ws` in the baseUrl to get the base subscription url 🤷‍♂ while it is a bit of a hack it is internal only. For users it probably makes the most sense

So yeah, I'm not sure what to go for here 😄 any opinion? We can also stick to the current implementation if the other approaches look worse

johannespfeiffer:
> So yeah, I'm not sure what to go for here 😄 any opinion? We can also stick to the current implementation if the other approaches look worse

Yes, let's stick with the current implementation. At the moment this "only" affects internal users. Maybe we will add some environment property (e.g. a enum with PROD, SANDBOX, DEV) when we have a sandbox.
LGTM 😄

PR #34, opened 2019-11-13
Extend tests; optimize test setup

johannespfeiffer:
This PR tries to optimize our test setup:

1. Removed `ts-node`, which created some problems during test transpilation and increased complexity.
2. Added some tests for Client, HttpRequest, Account and User.
3. Some general minor cleanup of the test setup.


package.json
```diff
@@ -2,38 +2,37 @@
...
-  "types": "dist/index.d.ts",
+  "main": "dist/lib/index.js",
+  "types": "dist/lib/index.d.ts",
   "repository": "https://github.com/kontist/js-sdk.git",
   "author": "Kontist GmbH",
   "license": "Apache-2.0",
   "scripts": {
     "build": "tsc && npx webpack",
-    "test": "mocha -r ts-node/register --recursive 'tests/**/*.spec.ts'",
+    "test": "mocha --recursive 'dist/tests/**/*.spec.js'",
```
hihuz: This works fine but only if we `npm run build`  before running the tests 🤔if `dist` is not present it will error, and if for example I am changing my tests while doing a feature I would need to manually `npm run build` before `npm t`.

Do you think we should do `"test": "npm run build && mocha --recursive 'dist/tests/**/*.spec.js'"` or something similar?

I might be missing something though 😄

lib/auth/tokenManager.ts
```diff
@@ -39,9 +39,9 @@ export class TokenManager {
 
     if (this.verifier) {
       // Implemented according to https://tools.ietf.org/html/rfc7636#appendix-A
-      const challenge = (
-        btoa(String.fromCharCode.apply(null, sha256.array(this.verifier))) || ""
-      )
+      const challenge = 
+        btoa(String.fromCharCode.apply(null, sha256.array(this.verifier)))!
```
hihuz: Oh cool, I didn't know about this operator 🙂 👍

package.json
johannespfeiffer: Good catch, thank you. Fixed in d4e6578a21a91f2ee53be14438574f35e1d596d4

PR #36, opened 2019-11-20
Add support for transfer fetching

hihuz:
Addresses: https://github.com/kontist/figure-backend/issues/5931

We want to add a way to fetch transfers of all types, keeping a consistent API with the existing transaction methods.

**Open questions / thoughts:**

* Should we keep the `for await (const transaction of client.models.transaction)` iterator on the base `iterableModel`, even if it is not usable for transfers? should we deprecate it? should we completely remove it (I am almost sure the webapp doesn't use it yet)? Removing it would obviously simplify the logic a little bit
* Is it fine to override the base `IterableModel` methods in `transfer` model to provide the correct argument type? I don't know if there is a better typescript way to do this
 
**Implementation details:**

Specifying the type for transfers is mandatory (we can only fetch transfers of a given type)

The current interface to fetch all transactions (`asyncIterator` property on the class) does not allow to pass any argument.

A possibility would have been to split the transfers models into 3 separate models, but I thought it was probably better to stick to the one model and change the existing interface a little bit.

So I implemented a method returning an asyncIterator:
```
  let transactions = [];
  for await (const transaction of client.models.transaction.fetchAll()) {
    transactions = transaction.concat(transaction);
  }
```

whereas before it was 
```
  let transactions = [];
  for await (const transaction of client.models.transaction) {
    transactions = transaction.concat(transaction);
  }
```

In order to avoid introducing any breaking changes, I updated the `IterableModel` to expose both.

For transfers, usage will be the same, but specifying `type` will be mandatory:
```
  let transfers = [];
  for await (const transfer of client.models.transfer.fetchAll({
    type: "SEPA_TRANSFER"
  })) {
    transfers = transfers.concat(transfer);
  }
```

Specifying type is also mandatory on the `fetch` method of transfer model:
```
const transfers = await client.models.transfer.fetch({
  type: "TIMED_ORDER"
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ An example to show how to fetch all user transactions
 
 ```typescript
 let transactions = [];
-for await (const transaction of client.models.transaction) {
+for await (const transaction of client.models.transaction.fetchAll()) {
   transactions = transactions.concat(transaction);
 }
 ```
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ const result = await client.graphQL.rawQuery(query);
 
 #### Transactions
 
-An example to show how to fetch all user transactions
+An example to show how to fetch all user transactions:
 
 ```typescript
 let transactions = [];
@@ -226,7 +226,7 @@ const result = await client.models.transfer.confirmMany(
 );
 ```
 
-An example to show how to fetch all user transfers of a given type
+An example to show how to fetch all user transfers of a given type:
 
 ```typescript
 let transfers = [];
--- a/README.md
+++ b/README.md
@@ -226,6 +226,25 @@ const result = await client.models.transfer.confirmMany(
 );
 ```
 
+An example to show how to fetch all user transfers of a given type
+
+```typescript
+let transfers = [];
+for await (const transfer of client.models.transfer.fetchAll({
+  type: Schema.TransferType.SepaTransfer
+})) {
+  transfers = transfers.concat(transfer);
+}
+```
+
+To fetch up to 50 latest transfers of a given type:
+
+```typescript
+const transfers = await client.models.transfer.fetch({
+  type: Schema.TransferType.TimedOrder
+});
+```
+
 ## MFA (Multi-Factor Authentication)
 
 Accessing Kontist banking APIs require Multi-Factor Authentication (MFA).
--- a/lib/graphql/iterableModel.ts
+++ b/lib/graphql/iterableModel.ts
@@ -1,19 +1,22 @@
 import { ResultPage } from "./resultPage";
 import { Model } from "./model";
+import { FetchOptions } from "./types";
 
 export abstract class IterableModel<T> extends Model<T> {
-  [Symbol.asyncIterator]() {
+  createAsyncIterator(args?: FetchOptions) {
+    const fetch = this.fetch.bind(this);
     let lastResult!: ResultPage<T>;
 
     return {
       next: async () => {
         // this is the first call or items are empty
         if (!lastResult) {
-          lastResult = await this.fetch();
+          lastResult = await fetch(args);
         }
 
         if (lastResult.items.length === 0 && lastResult.pageInfo.hasNextPage) {
-          lastResult = await this.fetch({
+          lastResult = await fetch({
+            ...args,
             after: lastResult.pageInfo.endCursor
           });
         }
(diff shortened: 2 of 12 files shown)


lib/graphql/iterableModel.ts
```diff
@@ -31,4 +34,17 @@ export abstract class IterableModel<T> extends Model<T> {
       }
     };
   }
+
+  [Symbol.asyncIterator]() {
```
hihuz: As mentioned in the PR description: 

Should we keep this iterator on the base `iterableModel`, even if it is not usable for transfers? should we deprecate it? should we completely remove it (I am almost sure the webapp doesn't use it yet)? Removing it would obviously simplify the logic a little bit

README.md
```diff
@@ -226,6 +226,25 @@ const result = await client.models.transfer.confirmMany(
 );
 ```
 
+An example to show how to fetch all user transfers of a given type:
+
+```typescript
+let transfers = [];
+for await (const transfer of client.models.transfer.fetchAll({
```
hihuz: Does this interface / namings look fine from a user's perspective?

lib/graphql/transfer.ts
```diff
@@ -203,7 +212,41 @@ export class Transfer extends Model<CreateTransferInput> {
...
+  createAsyncIterator(args: TransferFetchOptions) {
+    return super.createAsyncIterator(args);
+  }
+
+  /**
+   * @override specify query parameter arguments for transfers
+   *
+   * @param args query parameters
+   */
+  fetchAll(args: TransferFetchOptions) {
```
hihuz: Is it fine to override the base `IterableModel` methods here just to provide the correct argument type? I don't know if there is a better typescript way to do this

lib/graphql/iterableModel.ts
johannespfeiffer: Let’s remove it there, will not be usable for transfers anyway as you already said.

lib/graphql/transfer.ts
johannespfeiffer: Is it possible to use generics here? Sth. like IterableModel<T>

lib/graphql/iterableModel.ts
hihuz: All right, I agree, removed here: 3886585a94da3caabd3a84dab3c7f7ed89c52b75

lib/graphql/transfer.ts
hihuz: Great idea, I gave it a shot in this commit: f9b3bc212d90ec8b70184cc7a7f8ff68c8ab4f4f

It is fine for the most part but I have a question, I was forced to cast like so in a couple of places:
https://github.com/kontist/js-sdk/blob/f9b3bc212d90ec8b70184cc7a7f8ff68c8ab4f4f/lib/graphql/iterableModel.ts#L21-L24
otherwise I get an infamous `... is assignable to the constraint of type 'FetchOptionsType', but 'FetchOptionsType' could be instantiated with a different subtype of constraint` which makes sense in a way but I could not find another way around it

lib/graphql/resultPage.ts
```diff
@@ -1,22 +1,33 @@
...
+        model.fetch({ ...args, after: pageInfo.endCursor } as FetchOptionsType);
     }
 
     if (pageInfo.hasPreviousPage) {
-      this.previousPage = () => model.fetch({ before: pageInfo.startCursor });
+      this.previousPage = () =>
+        model.fetch({
+          ...args,
+          before: pageInfo.startCursor
+        } as FetchOptionsType);
```
johannespfeiffer: nit: can we skip this `as FetchOptionsType` here and at two other places? `args?` must not be optional for that to satisfy `tsc`.

lib/graphql/transfer.ts
johannespfeiffer: see https://github.com/kontist/js-sdk/pull/36#discussion_r350395494

TypeScript is right here, but message is not very comprehensible imho: Reason for this seems that `args` is allowed to be optional and then `{ ...args, after: lastResult.pageInfo.endCursor }`no longer satisfies `FetchOptionsType` (but has to because of `model: IFetch<ModelType, FetchOptionsType>,`) which could be set to e.g. `{ mandatoryProperty: string }` which would be missing then.
Hope this makes sense ;-) Let us discuss if you want to.

hihuz: Hmm it does absolutely make sense yes, but if I make `args` mandatory, I have to make it all the way up to `fetchAll(args: FetchOptionsType)`.

However for transactions, `args` has no mandatory field, so this would force users to go
`for await (const transaction of client.models.transaction.fetchAll({}))` (empty object argument) which is not so great 🤔 Or maybe is there a way to prevent this that I don't see? (apart from overriding `fetchAll` in transaction model, I don't see it 🙁)

hihuz: An illustration of what I mean can be seen in this commit: 2a704ffd76842829bfccfa39a09ca1a6c5ff663f 🙂

PR #37, opened 2019-11-21
Add enum for transaction category

hihuz:
Addresses: https://github.com/kontist/figure-backend/issues/5957

Just updating the part of the Schema relevant to the new transaction categories enum

**Note:** this is a *breaking* change. Webapp squad will need to be informed of this change when we release the SDK version containing these changes.

Corresponding backend PR: https://github.com/kontist/figure-backend/pull/6006

### DIFF
--- a/lib/graphql/schema.ts
+++ b/lib/graphql/schema.ts
@@ -624,7 +624,7 @@ export type Transaction = {
   directDebitFees: Array<DirectDebitFee>,
   name?: Maybe<Scalars['String']>,
   paymentMethod: Scalars['String'],
-  category?: Maybe<Scalars['String']>,
+  category?: Maybe<TransactionCategory>,
   userSelectedBookingDate?: Maybe<Scalars['DateTime']>,
   purpose?: Maybe<Scalars['String']>,
   documentNumber?: Maybe<Scalars['String']>,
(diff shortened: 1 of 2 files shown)


lib/graphql/schema.ts
```diff
@@ -635,6 +635,18 @@ export type Transaction = {
   originalAmount?: Maybe<Scalars['Int']>,
 };
 
+export enum TransactionCategory {
```
johannespfeiffer: I am a bit confused regarding the values here. Usually we have them all uppercase. Do you know a reason for this?

hihuz: Yeah, I had a specific reason for it, initially I put them uppercase but this was introducing a breaking API change (as opposed to just a SDK version breaking change, this one would affect every user as soon as the backend is deployed). If the enum `keys` are uppercased in our backend, this is what will be exposed by GraphQL, e.g. it would return
`category: "VAT0"` whereas currently it returns `category: "vat0"`

If we go ahead and do this change to uppercase, it will make rolling out the change very painful, because it would break any consumer code using transactions, and the breakage would occur as soon as we release the backend. This is would not just affect SDK users but all users of our GraphQL API.

Just to guarantee that it would work for the web app, we would need to:
* prepare the backend PR doing the change and wait to merge it
* prepare a SDK version including the change and release it (it would be broken until the backend change goes through)
* prepare a webapp PR including the update to the new SDK version containing the change
* release the backend in prod with the change
* release the webapp in prod with the change

Last two points would need to happen exactly at the same time, and also we would need to release a SDK version that would be not working until all is done.

Also the above does not account for any potential external user of our GQL api (money money? anybody else?), they would have to figure out the change by themselves and update accordingly

So I took the way easier path of just keeping the enum keys the same as they were until now.

But I can definitely understand the argument for consistency, if you think it's a better choice I can definitely do it and I'll get someone from websquad to sync this whole thing

hihuz: One argument for consistency is I don't see the webapp using gql transactions in production right now, but I do know that they will soon as I have seen the code on develop

PR #39, opened 2019-11-27
Fix async iterator return order

hihuz:
While writing some tests for the SDK on the backend, I noticed that the asyncIterator returned results in the reverse order for each 'batch' of items.

e.g. if I have 2 transactions first with booking date 2019-11-27 and second with booking date 2019-11-26, regular `fetch()` will return them in this order: `["2019-11-27", "2019-11-26"]` which is the order our GQL api also returns them by default (latest transactions always first), but `fetchAll()` will return them in the opposite order `["2019-11-26", "2019-11-27"]. I believe this is incorrect.

This was due to the fact that we used the `pop()` which will take the last item in the array to return each item in the async iterator logic.

We can simply change it to `shift()` to fix the issue.

Also our existing tests didn't demonstrate the issue because we only had pages with 1 item each, so I added a test with multiple items per page (and it would fail with previous `pop()` logic)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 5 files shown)


hihuz:
Just a sec before reviewing, I forgot to add the same tests for transfers 🙈

PR #41, opened 2019-12-17
Confirmed refresh tokens

hihuz:
Addresses: https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-backend/5927 (at least some parts of it 😅)

We will now be issuing confirmed refresh tokens when passing mfa (push notif or device binding) with offline scope: https://github.com/kontist/figure-backend/pull/6110

We need to update our sdk to also store and return this refresh token so that our sdk users can benefit from this feature.

Will update documentation next


lib/utils/extractTokensFromMfaResponse.ts
```diff
@@ -0,0 +1,21 @@
...
+): ConfirmedTokenData => {
+  const { token: accessToken, refresh_token: confirmedRefreshToken } = mfaResponse;
+
+  const { refreshToken: existingRefreshToken } = currentToken || {};
+
+  return {
+    accessToken,
+    refreshToken: confirmedRefreshToken || existingRefreshToken
+  };
+};
```
johannespfeiffer: Might be possible to simplify this a bit:
`this.tokenManager.setToken` allows the 2nd parameter to be `undefined` and just sets the access token in that case.
So we do not need the `currentToken` parameter here and can just return `mfaResponse.refresh_token` which might be `undefined` but which is okay. Maybe we do not even need this method at all. Or do I miss sth. here?

hihuz: Yeah, I knew about the optional second param, the case I was trying to handle here was:
* I have saved a refresh token on my sdk instance (whatever the origin / source of it)
* I do MFA with push or device binding, and API returns me only an access token and no refresh token
* I figure I can still keep the refresh token I had saved from before instead of deleting it

because otherwise refresh token will be deleted in `setToken` method (we create a new 'token' instance without refresh token): 
https://github.com/kontist/js-sdk/blob/b79206a4ff52a56c9f366f30c11364929d72bf61/lib/auth/tokenManager.ts#L186-L192

I'm not sure if this case could even happen in practice though, maybe it's a bit of a stretch to try to accommodate for it? code would indeed be way simpler if we just didn't care about this above case I described.

Also, initially in the 'device' logic, we would keep the previous 'non-confirmed' refresh token (although arguably the case is a bit different now):
https://github.com/kontist/js-sdk/blob/b79206a4ff52a56c9f366f30c11364929d72bf61/lib/auth/device.ts#L89


What do you think? I'm fine with ditching all of this honestly if you think it's too much for just an edge case

hihuz: So after discussing the issue, we agreed with @johannespfeiffer that we would remove all of this logic for now, it simplifies the overall implementation quite a bit and only handles an edge case which doesn't seem to be valid from a user's perspective anyway.

Have removed it here: 6ebcab3074d0eb968bb7d9051670f38d21d3a6ac

PR #42, opened 2019-12-18
Add card model and fetching methods

hihuz:

Let's proceed by steps and introduce first the model and fetching method for cards, have tested locally and works fine.

Please note that on this feature branch the build is not .gitignored but this is on purpose (this will be necessary if we want to get access to SDK card methods in native app without doing a SDK release)

### DIFF
--- a/lib/graphql/card.ts
+++ b/lib/graphql/card.ts
@@ -1,9 +1,7 @@
 import { Card as CardModel, Query } from "./schema";
 import { Model } from "./model";
 import { ResultPage } from "./resultPage";
-import { KontistSDKError } from "../errors";
-import { CardOptions } from "./types"
-
+import { CardOptions } from "./types";
 
 const CARD_FIELDS = `
   id
@@ -71,21 +69,23 @@ const GET_CARD = `
 `;
 
 export class Card extends Model<CardModel> {
-  async fetch(): Promise<ResultPage<CardModel>> {
-    throw new KontistSDKError({
-      message:
-        "Card model does not implement fetch, please use the `get` method instead."
-    });
-  }
-
-  /**
-   * Returns details of all cards belonging to the current user
+    /**
+   * Fetches all cards belonging to the current user
    *
-   * @returns details of the cards
+   * @returns     result page
    */
-  async getAll(): Promise<Array<CardModel> | null> {
+  async fetch(): Promise<ResultPage<CardModel>> {
     const result: Query = await this.client.rawQuery(GET_CARDS);
-    return result.viewer.mainAccount?.cards ?? null;
+
+    const cards = result.viewer.mainAccount?.cards ?? [];
+
+    // cards resolver is not paginated
+    const pageInfo = {
+      hasNextPage: false,
+      hasPreviousPage: false
+    };
+
+    return new ResultPage(this, cards, pageInfo, {});
   }
 
   /**
--- a/lib/graphql/card.ts
+++ b/lib/graphql/card.ts
@@ -7,34 +7,11 @@ const CARD_FIELDS = `
   id
   status
   type
-  canceledAt
-  expirationDate
-  orderedAt
   holder
   formattedExpirationDate
   maskedPan
   settings {
     contactlessEnabled
-    cardNotPresentLimits {
-      daily {
-        maxAmountCents
-        maxTransactions
-      }
-      monthly {
-        maxAmountCents
-        maxTransactions
-      }
-    }
-    cardPresentLimits {
-      daily {
-        maxAmountCents
-        maxTransactions
-      }
-      monthly {
-        maxAmountCents
-        maxTransactions
-      }
-    }
   }
 `;
 
--- a/lib/graphql/card.ts
+++ b/lib/graphql/card.ts
@@ -1,0 +1,101 @@
+import { Card as CardModel, Query } from "./schema";
+import { Model } from "./model";
+import { ResultPage } from "./resultPage";
+import { KontistSDKError } from "../errors";
+import { CardOptions } from "./types"
+
+
+const CARD_FIELDS = `
+  id
+  status
+  type
+  canceledAt
+  expirationDate
+  orderedAt
+  holder
+  formattedExpirationDate
+  maskedPan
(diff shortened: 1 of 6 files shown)


lib/graphql/card.ts
```diff
@@ -0,0 +1,101 @@
...
+        ) {
+          ${CARD_FIELDS}
+        }
+      }
+    }
+  }
+`;
+
+export class Card extends Model<CardModel> {
+  async fetch(): Promise<ResultPage<CardModel>> {
```
hihuz: I wanted to use `fetch` to get all cards of user, but unfortunately it needs to implement `ResultPage<ModelType, FetchOptionsType>` as per our base model.

So I went with a different method, `getAll()` and `get({})`, how does it look? As user of this interface @p-janik let me know if you would like anything different or have better ideas

lib/graphql/card.ts
```diff
@@ -0,0 +1,101 @@
+import { Card as CardModel, Query } from "./schema";
+import { Model } from "./model";
+import { ResultPage } from "./resultPage";
+import { KontistSDKError } from "../errors";
+import { CardOptions } from "./types"
+
+
+const CARD_FIELDS = `
```
hihuz: should we expose all of these fields in SDK? I don't know much about the business logic behind these so I just exposed everything (cc @p-janik 😛)

lib/graphql/card.ts
johannespfeiffer: Did not get it - why is it not possible to implement that interface?

hihuz: cards resolver is not paginated, GQL returns all cards, so there is no notion of cursors, edge, etc.

hihuz: hmm now taht I think about it we could still use it, just need to provide 
```
      hasNextPage: false,
      hasPreviousPage: false
```

every time without cursors 🤔 let me update

hihuz: any idea about the method name to get card by ID? just `get({ ... })` ?

hihuz: Here: ab2bcb0ca0155a5d2bfa1b46cf0163024702a7e5

Thank you for your comment 🙏

lib/graphql/card.ts
```diff
@@ -0,0 +1,101 @@
...
+    return new ResultPage(this, cards, pageInfo, {});
+  }
+
+  /**
+   * Returns details of a specific card belonging to the current user
+   *
+   * @param args  query parameters
+   * @returns     details of the card specified in query parameters
+   */
+  async get(args: CardOptions): Promise<CardModel | null> {
```
hihuz: Any other idea for the name of this method?

lib/graphql/card.ts
p-janik: Not all of them have to be exposed,  will double-check those values 👍

tests/graphql/card.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        cardId: cardData.id,
+        type: CardType.VisaBusinessDebit
+      });
+
+      // assert
+      sinon.assert.calledOnce(spyOnRawQuery);
+      expect(result).to.deep.eq(cardData);
+    });
+
+    it("should call rawQuery and return null for missing account", async () => {
```
p-janik: `missing card`, right?

lib/graphql/card.ts
p-janik: hard to find one.. maybe keep this since we have the same name for user, account

lib/graphql/card.ts
```diff
@@ -0,0 +1,101 @@
...
+  type
+  canceledAt
+  expirationDate
+  orderedAt
+  holder
+  formattedExpirationDate
+  maskedPan
+  settings {
+    contactlessEnabled
+    cardNotPresentLimits {
```
p-janik: I'm not sure about exposing cardLimits directly here, this would trigger 2 calls from Kontist backend to Solaris (one for cardNotPresentLimits and one for cardPresentLimits). Limits will be displayed on a separate screen and doesn't have to be fetched by default

tests/graphql/card.spec.ts
hihuz: Hmm it is really testing when account is missing though, e.g.: `resolves({ viewer: {} })`

I wanted to test that sdk is not crashing in that case because it can happen that an sdk user checks cards for an end user which doesn't have account created yet, if it makes sense? 🤔 

If user has missing cards (no cards) gql returns empty array already, so I don't really need to test for that specifically I think

lib/graphql/card.ts
hihuz: so, do you think I should add them for example to `get({})` method only? (the one who gets one card by ID), and generic `fetch()` method would not return them, does it sound correct?

lib/graphql/card.ts
hihuz: yeah, that's why I chose it in the first place, will be fine 👍

lib/graphql/card.ts
hihuz: or maybe we will have another method specifically for these limits later? like `client.models.card.getLimits({ cardId })`?

I think I will remove these limits from this PR for now so that it becomes mergeable, and we can adjust code later on to expose limits somewhere

lib/graphql/card.ts
hihuz: Have removed the ones we agreed upon in this commit: 08bd85269213474117035208c67d728dbb60d883

lib/graphql/card.ts
hihuz: So, for now I removed the limits entirely, we will see how we decide to introduce them alter on.

Here it is: 08bd85269213474117035208c67d728dbb60d883

tests/graphql/card.spec.ts
p-janik: got it, you re right :)

lib/graphql/card.ts
p-janik: your solution looks good to me, we can extend SDK with new method when we will be close to work with card limit screens :)

PR #43, opened 2019-12-18
Add method to activate cards

hihuz:

In this PR I am adding method to activate cards, tested locally and works fine

### DIFF
--- a/lib/graphql/card.ts
+++ b/lib/graphql/card.ts
@@ -1,7 +1,7 @@
 import { Card as CardModel, Query } from "./schema";
 import { Model } from "./model";
 import { ResultPage } from "./resultPage";
-import { CardOptions } from "./types";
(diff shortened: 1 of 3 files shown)


lib/graphql/card.ts
```diff
@@ -94,8 +106,19 @@ export class Card extends Model<CardModel> {
...
     return result.viewer.mainAccount?.card ?? null;
   }
+
+  /**
+   * Activates a card
+   *
+   * @param args  query parameters
+   * @returns     activated card
+   */
+  async activate(args: ActivateCardOptions): Promise<CardModel> {
```
hihuz: should I do anything to handle errors here? or should we leave that to the consumer / end user?

p-janik: let's leave it for end-user, at least for now

PR #46, opened 2019-12-19
Add helper to create cards

hihuz:

Adding helper to create cards, have one doubt about interface here, let me know 🙂

### DIFF
--- a/lib/graphql/card.ts
+++ b/lib/graphql/card.ts
@@ -54,10 +54,10 @@ const GET_CARD = `
 `;
 
 const CREATE_CARD = `mutation createCard(
-  $cardType: CardType!
+  $type: CardType!
 ) {
   createCard(
-    cardType: $cardType
+    type: $type
   ) {
     ${CARD_FIELDS}
   }
--- a/lib/graphql/card.ts
+++ b/lib/graphql/card.ts
@@ -6,7 +6,8 @@ import {
   ActivateCardOptions,
   ChangeCardPINOptions,
   ConfirmChangeCardPINOptions,
-  ChangeCardStatusOptions
+  ChangeCardStatusOptions,
+  CreateCardOptions
 } from "./types";
 
 const CARD_FIELDS = `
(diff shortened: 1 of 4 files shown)


lib/graphql/card.ts
```diff
@@ -130,11 +140,22 @@ export class Card extends Model<CardModel> {
     return result.viewer.mainAccount?.card ?? null;
   }
 
+  /**
+   * Creates a card
+   *
+   * @param cardType   the type of card to create
+   * @returns          the newly created card details
+   */
+  async create(cardType: CardType): Promise<CardModel> {
```
hihuz: For all other card methods for now I have used `args` object argument, but in the case of card creation there is only one argument, should I still use object argument for consistency? or should I use cardType directly like I've done here?

I'm not sure 🤔

johannespfeiffer: Yes, that looks like a good idea to me. By using `args` instead of `cardType` we can stay backward compatible if we want to extend the options in the future (e.g. with the PIN)

johannespfeiffer: And I think we can apply the same logic to `cardType` -> `type` as we did to `cardId` -> `id` and just use `type` here, any objections?

hihuz: Great point about future extensions 👍 Will change it. Regarding `cardType` vs `type` I'd also agree 🤔I'll prepare a backend PR and propose it to the card squad

hihuz: changed to options argument here: 69514635633c9957957aabe0034ff54f8882253f

renamed argument to `type` here: e696825b8fe32928f0fa7aa79866ae3a2c6c9f24

and here is the backend PR doing the name change: https://github.com/kontist/figure-backend/pull/6171

lib/graphql/card.ts
```diff
@@ -52,6 +53,16 @@ const GET_CARD = `
   }
 `;
 
+const CREATE_CARD = `mutation createCard(
+  $cardType: CardType!
+) {
+  createCard(
+    cardType: $cardType
```
johannespfeiffer: I was under the impression that we changed that in the backend?

hihuz: You are absolutely correct, thank you for spotting it.

I forgot to update the query and it's not the first time 😬 will be more careful about it next time.

I have updated it and double checked that it works against current develop backend now: d7bf808bc9559770462e6470d9703637cc93aa85

PR #49, opened 2019-12-20
Add card settings helper

hihuz:

Adding card settings helper.

I'm unsure about my approach here. We could also have separate methods to update `contactlessEnabled` and `limits` 🤷‍♂ let me know what you think


hihuz:
Will add a commit to update build before merging the PR

PR #58, opened 2020-01-14
Add solaris cards methods

hihuz:
Let's merge and release these 😛 

All of these commits were approved in their respective PRs


hihuz:
🚀

PR #61, opened 2020-01-21
Add card replace method

goodfellamilos:
Addresses: [4616](https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-app/4616)

Corresponding backend PR: [6265](https://github.com/kontist/figure-backend/pull/6265)

### DIFF
--- a/lib/graphql/card.ts
+++ b/lib/graphql/card.ts
@@ -301,7 +301,7 @@ export class Card extends Model<CardModel> {
    * @param args   query parameters including card id
    * @returns      the newly created card details
    */
-  async replaceCard(args: ReplaceCardOptions): Promise<CardModel> {
+  async replace(args: ReplaceCardOptions): Promise<CardModel> {
     const result = await this.client.rawQuery(REPLACE_CARD, args);
     return result.replaceCard;
   }
--- a/lib/graphql/card.ts
+++ b/lib/graphql/card.ts
@@ -8,7 +8,8 @@ import {
   ConfirmChangeCardPINOptions,
   ChangeCardStatusOptions,
   CreateCardOptions,
-  UpdateCardSettingsOptions
+  UpdateCardSettingsOptions,
+  ReplaceCardOptions
 } from "./types";
 
 const CARD_FIELDS = `
@@ -176,6 +177,16 @@ const UPDATE_CARD_SETTINGS = `mutation updateCardSettings(
   }
 }`;
 
+const REPLACE_CARD = `mutation replaceCard(
+  $id: String!
+) {
+  replaceCard(
+    id: $id
+  ) {
+    ${CARD_FIELDS}
+  }
+}`;
+
 export class Card extends Model<CardModel> {
   /**
    * Fetches all cards belonging to the current user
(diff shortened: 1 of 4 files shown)


lib/graphql/schema.ts
```diff
@@ -367,6 +367,7 @@ export type Mutation = {
   changeCardStatus: Card,
   changeCardPIN: ConfirmationRequest,
   confirmChangeCardPIN: ConfirmationStatus,
+  replaceCard: Card,
```
hihuz: Did you change the schema manually? We have a specific procedure to update it automatically from backend, sorry that it isn't documented, you can:
1) in backend, spin it up and comment in `services/backendService/src/core/graphql/server.ts`:
```
export const createGraphQLServer = async () => {
  const graphQLServer = express.Router();
  // graphQLServer.use(middlewares.getAuth());
  // graphQLServer.use(middlewares.extendRequestWithUser);
  graphQLServer.use(
    graphqlHTTP({
      schema: await createGraphQLSchema()
    })
  );
  return graphQLServer;
};
```
(this is to allow access to the schema without being authenticated)

2) in sdk run: `npm run schema:generate` 

3) commit corresponding changes to the schema

For example, when I do it, I can see that we are missing:
```
export type MutationReplaceCardArgs = {
  id: Scalars['String']
};
```

There are also some added stuff for `reorderCard`, you can also commit them if you like or leave them for the PR who will include the helper for that.

lib/graphql/card.ts
```diff
@@ -283,4 +294,15 @@ export class Card extends Model<CardModel> {
...
     return result.updateCardSettings;
   }
+
+  /**
+   * Replace a card
+   *
+   * @param args   query parameters including card id
+   * @returns      the newly created card details
+   */
+  async replaceCard(args: ReplaceCardOptions): Promise<CardModel> {
```
hihuz: what do you think of renaming this method just `replace`? otherwise usage would be `kontistClient.card.replaceCard` mentioning "card" twice is a bit redundant, also for other methods we took the same approach (e.g. `kontistClient.card.create`, etc.)

lib/graphql/schema.ts
hihuz: Also, a good remark from @johannespfeiffer once the schema is updated with this new type (`MutationReplaceCardArgs`) as described in my comment above, it would make sense to use that directly as the type for the `replace` method (e.g. `async replace(args: MutationReplaceCardArgs) { ... }`) and get rid of the `ReplaceCardOptions` type in `lib/graphql/types` entirely.

I know that you followed the pattern used for other card helpers but as @johannespfeiffer correctly pointed out, it was a mistake from us

Please let me know if this is not clear 🙂

lib/graphql/card.ts
goodfellamilos: Yeah, I has this idea after I already pushed it :) Especially since next method I need to implement is `reorderCard (reorder)`. It totally makes sense, will update it 👍

hihuz:
Oh, almost forgot: could you also add an entry in the `CHANGELOG.md` ?

under 
```
## [Unreleased]
### Added
```

goodfellamilos:
> The id for the new card is the same as the old one when I do a replacement, but this is most likely an issue with our mock and has nothing to do with the SDK helper in any case.

I think this is not an issue, the logic is that we just update the existing card in db with new status, while card id remains the same. 

Check https://github.com/kontist/figure-backend/blob/45dcc2cc42adb307c16c0543f72cf3c56c1a8f6c/services/backendService/src/modules/creditcard/integrations/solaris/card/api/replaceCard.ts

hihuz:
Makes sense 🙂 I mentioned it just in case

PR #62, opened 2020-01-21
Use schema types instead manual ones

johannespfeiffer:
We recently found a few manual types which could be replaced by existing autogenerated schema types. We should use them instead to avoid code duplication and manual work.


johannespfeiffer:
> Looks great, thank you and sorry for this mixup 🙂
> 
> One small thing, in `lib/graphql/types.ts` we now have 2 unused imports: `CardAction` and `CardLimitsInput` from "./schema" what do you think about removing them?

Makes sense. To avoid things like that in the future I did create https://github.com/kontist/js-sdk/pull/64

PR #63, opened 2020-01-21
Add card reorder method

goodfellamilos:
Addresses: [4616](https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-app/4616)

Corresponding backend PR: [6265](https://github.com/kontist/figure-backend/pull/6267)

`reorder` is different from `replace` method introduced in #61 in a way that `reorder` makes 2 calls to solaris (close current card + create new card) unlike `replace` which calls solaris replace endpoint once (`/cards/:card_id/replace`) and only updates existing card in db.

`reorder` method is called when customer's card is damaged.
`replace` method is called when customer's card is lost or stolen.


johannespfeiffer:
> `reorder` method is called when customer's card is damaged.
> `replace` method is called when customer's card is lost or stolen.

Great explanation - will steal it and add it as description to the mutations 😄

goodfellamilos:
Awesome, thanks @johannespfeiffer 🙌 Merging it then :)

PR #64, opened 2020-01-21
Apply basic linting rules

johannespfeiffer:
This PR adds a very basic tslint config.
Instead of arguing about good or bad rules we should just stick to the default values and adjust where required (see tslint.json).

### DIFF
--- a/lib/graphql/transaction.ts
+++ b/lib/graphql/transaction.ts
@@ -87,7 +87,7 @@ export class Transaction extends IterableModel<TransactionModel> {
     const pageInfo = result?.viewer?.mainAccount?.transactions?.pageInfo ?? {
       hasNextPage: false,
       hasPreviousPage: false,
-    }           ;
+    };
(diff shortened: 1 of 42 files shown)


lib/graphql/transaction.ts
```diff
@@ -86,27 +86,27 @@ export class Transaction extends IterableModel<TransactionModel> {
 
     const pageInfo = result?.viewer?.mainAccount?.transactions?.pageInfo ?? {
       hasNextPage: false,
-      hasPreviousPage: false
-    };
+      hasPreviousPage: false,
+    }           ;
```
hihuz: This looks a bit weird 🤔 is it a typo or something?

johannespfeiffer: Good catch! Yes - initially I used an old version of tslint which did not knew about the used nullish coalescing operator above and it created that silly whitespace.
Found and fixed one more occurrence.

PR #65, opened 2020-01-22
Update changelog

goodfellamilos:
Card replace and reorder


goodfellamilos:
I added changes under `Unreleased`. I don't know at which point do we set a date there. I can update it here if that's how it should be done.

hihuz:
Nah it's good like this 🙂Usually we update it with number and date after the release actually happened, in case other stuff are included, speaking of which I will try to include this as well https://github.com/kontist/js-sdk/pull/60

Let's merge your PR first and then I'll rebase master in mine 🙂

PR #69, opened 2020-02-04
Update schema

hihuz:
The previous version of type-graphql was exposing some non-necessary types.

I'll double check that nothing is breaking with this update but I don't expect it.


hihuz:
Closing in favor of https://github.com/kontist/js-sdk/pull/72

PR #71, opened 2020-02-14
Add filtering to transactions

hihuz:
Addresses: https://github.com/kontist/figure-backend/issues/6151

Corresponding backend PR: https://github.com/kontist/figure-backend/pull/6409

So I went with extending the fetch method to allow filter, and also added a "search" helper which does some formatting on a passed string to cover the most generic use case (the one we will need to implement in both webapp and native app).

I think it's nice to have this additional helper for unifying logic in both our apps, and also provide some easy to use method if users don't want to dig too deep into filters

### DIFF
--- a/lib/graphql/transaction.ts
+++ b/lib/graphql/transaction.ts
@@ -31,25 +31,27 @@ const TRANSACTION_FIELDS = `
   documentType
 `;
 
-const FETCH_TRANSACTIONS = `query fetchTransactions ($first: Int, $last: Int, $after: String, $before: String) {
-  viewer {
-    mainAccount {
-      transactions(first: $first, last: $last, after: $after, before: $before) {
-        edges {
-          node {
-            ${TRANSACTION_FIELDS}
+const FETCH_TRANSACTIONS = `
+  query fetchTransactions ($first: Int, $last: Int, $after: String, $before: String, $filter: TransactionFilter) {
+    viewer {
+      mainAccount {
+        transactions(first: $first, last: $last, after: $after, before: $before, filter: $filter) {
+          edges {
+            node {
+              ${TRANSACTION_FIELDS}
+            }
           }
-        }
-        pageInfo {
-          hasNextPage
-          hasPreviousPage
-          startCursor
-          endCursor
+          pageInfo {
+            hasNextPage
+            hasPreviousPage
+            startCursor
+            endCursor
+          }
         }
       }
     }
   }
-}`;
+`;
 
 export const NEW_TRANSACTION_SUBSCRIPTION = `subscription {
   newTransaction {
--- a/lib/graphql/transaction.ts
+++ b/lib/graphql/transaction.ts
@@ -155,7 +155,7 @@ export class Transaction extends IterableModel<TransactionModel> {
     const amountTerms = searchTerms
       .filter((term) => amountRegex.test(term))
       .reduce((terms: number[], term: string): number[] => {
-        const amountInCents = parseFloat(term.replace(",", ".")) * 100;
+        const amountInCents = Math.round(parseFloat(term.replace(",", ".")) * 100);
         return [...terms, amountInCents, amountInCents * -1];
       } , []);
 
--- a/lib/graphql/transaction.ts
+++ b/lib/graphql/transaction.ts
@@ -1,6 +1,7 @@
 import { IterableModel } from "./iterableModel";
 import { ResultPage } from "./resultPage";
 import {
+  AccountTransactionsArgs,
   MutationCategorizeTransactionArgs,
   Query,
   Transaction as TransactionModel,
@@ -77,7 +78,7 @@ export class Transaction extends IterableModel<TransactionModel> {
    * @param args  query parameters
    * @returns     result page
    */
-  public async fetch(args?: FetchOptions): Promise<ResultPage<TransactionModel>> {
+  public async fetch(args?: AccountTransactionsArgs): Promise<ResultPage<TransactionModel>> {
     const result: Query = await this.client.rawQuery(FETCH_TRANSACTIONS, args);
 
     const transactions = (
--- a/lib/graphql/transaction.ts
+++ b/lib/graphql/transaction.ts
@@ -2,12 +2,16 @@ import { IterableModel } from "./iterableModel";
 import { ResultPage } from "./resultPage";
 import {
   AccountTransactionsArgs,
+  BaseOperator,
   MutationCategorizeTransactionArgs,
   Query,
   Transaction as TransactionModel,
+  TransactionFilter,
   TransactionsConnectionEdge,
 } from "./schema";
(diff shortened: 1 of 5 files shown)


lib/graphql/transaction.ts
```diff
@@ -1,13 +1,18 @@
...
+  BaseOperator,
   MutationCategorizeTransactionArgs,
   Query,
   Transaction as TransactionModel,
+  TransactionFilter,
   TransactionsConnectionEdge,
 } from "./schema";
 import { FetchOptions, Subscription, SubscriptionType } from "./types";
 
+const MAX_SEARCH_QUERY_LENGTH = 200;
```
hihuz: What do you think of truncating user query if too long? is 200 chars proper?

Do you also think I should perform some other sanitization operations before sending the stuff to the backend? 🤔

lib/graphql/transaction.ts
```diff
@@ -120,4 +140,28 @@ export class Transaction extends IterableModel<TransactionModel> {
...
+    const searchTerms = searchQuery.slice(0, MAX_SEARCH_QUERY_LENGTH).split(" ");
+
+    const filter: TransactionFilter = {
+      iban_likeAny: searchTerms,
+      name_likeAny: searchTerms,
+      operator: BaseOperator.Or,
+      purpose_likeAny: searchTerms,
+    };
+
+    const amountRegex = /^-?\d+([,.]\d{1,2})?$/;
```
hihuz: So, I want to match: `1234`, `-1234`, `1.23`, `12.1`, `1,23`, `12,34` etc. etc.

I don't want to match shit like `.10`, `10.`, `10,` but maybe I should? 🤔

lib/graphql/types.ts
```diff
@@ -1,22 +1,12 @@
...
-  last?: number;
-  before?: string | null;
-  after?: string | null;
-};
-
-export type TransferFetchOptions = {
-  type: TransferType;
-  where?: TransfersConnectionFilter;
-} & FetchOptions;
+export type FetchOptions = AccountTransactionsArgs | AccountTransfersArgs;
```
hihuz: what do you think of this? 🤔 it's not strictly needed but we were once again creating some custom types that we can just get from the Schema

lib/graphql/transaction.ts
p-janik: I would vote for obvious ones only, so no .10 💩

lib/graphql/transaction.ts
p-janik: 200 I a lot and should be enough, I don't think that user would type that much or paste that long strings

lib/graphql/transaction.ts
```diff
@@ -91,6 +98,19 @@ export class Transaction extends IterableModel<TransactionModel> {
     return new ResultPage(this, transactions, pageInfo, args);
   }
 
+  /**
+   * Fetches first 50 matching transactions for a given user input
```
p-janik: `fetches up to first 50`?

lib/graphql/transaction.ts
```diff
@@ -120,4 +140,28 @@ export class Transaction extends IterableModel<TransactionModel> {
...
+      name_likeAny: searchTerms,
+      operator: BaseOperator.Or,
+      purpose_likeAny: searchTerms,
+    };
+
+    const amountRegex = /^-?\d+([,.]\d{1,2})?$/;
+    const amountTerms = searchTerms
+      .filter((term) => amountRegex.test(term))
+      .reduce((terms: number[], term: string): number[] => {
+        const amountInCents = parseFloat(term.replace(",", ".")) * 100;
```
p-janik: what will happen if i provide `33.91`? :P

hihuz: oh shit nice catch 🤦‍♂ what if I Math.round afterwards? it should work for all cases, I'm trying to come up with a counter example

lib/graphql/types.ts
johannespfeiffer: Not a fan of this change (`FetchOptions`):
1. We will need to adjust this for every new collection.
2. It allows using `transaction.fetchAll` with `args` of type `AccountTransfersArgs`.

I do like that we can get rid of `TransferFetchOptions`.

lib/graphql/transaction.ts
johannespfeiffer: 200 characters should be enough.

lib/graphql/transaction.ts
```diff
@@ -91,6 +98,19 @@ export class Transaction extends IterableModel<TransactionModel> {
     return new ResultPage(this, transactions, pageInfo, args);
   }
 
+  /**
+   * Fetches first 50 matching transactions for a given user input
+   * It will consider case insentsitive like matches for amount,
```
johannespfeiffer: nit: typo "insentsitive"

tests/graphql/transaction.spec.ts
```diff
@@ -75,6 +81,29 @@ describe("Transaction", () => {
...
+          operator: BaseOperator.And,
+          amount_gt: 1000,
+          name_like:"SaNtA"
+        }
+      });
+
+      expect(results.items).to.deep.equal([
+        firstTransaction,
+        thirdTransaction,
+      ]);
```
johannespfeiffer: Can we extend the expectation a bit? E.g.
```
      sinon.assert.calledWithExactly(stub, sinon.match.string, {
        filter: {
          operator: BaseOperator.And,
          amount_gt: 1001,
          name_like: "SaNtA"
        }
      });
```

Without it the test runs fine even without the filter.

lib/graphql/transaction.ts
hihuz: I think `Math.round` should work in all cases: 84d001fa0a21b1e74b87c5e99108e8a05d855c7a

lib/graphql/types.ts
hihuz: Makes total sense, I reverted that part of my changes: 0032192b41d45d1f6fd152fc7e70fa7bc1768709

lib/graphql/transaction.ts
hihuz: Thank you 🙂 c8d7e7dec33255d94d7e49a1b6c3d54ac3a484f9

lib/graphql/transaction.ts
hihuz: Good idea, c8d7e7dec33255d94d7e49a1b6c3d54ac3a484f9

tests/graphql/transaction.spec.ts
hihuz: Um good point 😂 test was pretty useless. What about this assertion 9aec03fb1bc6c0f27a521c4f75815ea354e38e1a ?

johannespfeiffer: 👍 thank you

PR #73, opened 2020-02-18
updateTransfer for categorization data

p-janik:
With this PR client can update categorization metadata of standing order, timed order and sepa transfer.

This is blocked on https://github.com/kontist/figure-backend/compare/develop...epic/future-transaction-categorization

### DIFF
--- a/lib/graphql/schema.ts
+++ b/lib/graphql/schema.ts
@@ -261,6 +261,10 @@ export type CreateTransferInput = {
   e2eId?: Maybe<Scalars['String']>,
   /** The reoccurrence type of the payments for Standing Orders */
   reoccurrence?: Maybe<StandingOrderReoccurenceType>,
+  /** The user selected category for the SEPA Transfer */
+  category?: Maybe<TransactionCategory>,
+  /** When a transaction corresponds to a tax or vat payment, the user may specify at which date it should be considered booked */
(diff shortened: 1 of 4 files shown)


lib/graphql/schema.ts
```diff
@@ -1060,14 +1068,14 @@ export type UpdateClientInput = {
...
+/** The available fields to update a transfer */
 export type UpdateTransferInput = {
-  /** The ID of the Standing Order to update */
+  /** The ID of the transfer to update */
   id: Scalars['String'],
   /** The type of transfer to update, currently only Standing Orders are supported */
-  type: TransferType,
-  /** The amount of each Standing Order payment in cents */
-  amount: Scalars['Int'],
+  type?: Maybe<TransferType>,
```
hihuz: It's not related to this PR but rather to the corresponding backend PR, I didn't notice that we changed this 🤔 do you remember the reason why we made it optional? It seems backend will throw if it is not provided

p-janik: huh, true, this should not be optional 🤔

hihuz: It's fine, we can adjust it later if you need to have this for front-end, it's not really a big issue 👍

p-janik: @hihuz I've created a PR to fix it on backend side: https://github.com/kontist/figure-backend/pull/6504

PR #77, opened 2020-03-03
Extend transfer fields

lera:
Add missing category data on fetching transfer

### DIFF
--- a/lib/graphql/transfer.ts
+++ b/lib/graphql/transfer.ts
@@ -24,8 +24,8 @@ const TRANSFER_FIELDS = `
   purpose
   e2eId
   reoccurrence
-  nextOccurrence,
-  category,
+  nextOccurrence
+  category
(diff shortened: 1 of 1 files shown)


lib/graphql/transfer.ts
```diff
@@ -24,7 +24,9 @@ const TRANSFER_FIELDS = `
   purpose
   e2eId
   reoccurrence
-  nextOccurrence
+  nextOccurrence,
```
hihuz: it would work with `,` but for consistency do you think we could remove them? Just line return is enough to separate fields

lera: 🤦‍♀️ didn't notice it

hihuz: Not a problem 🙂 We have also not been always consistent with this across the repo I think

PR #78, opened 2020-03-03
Match wild cent amounts when appropriate in search

hihuz:
While testing search, Dani had the following idea regarding amount search:
> regarding the amounts I thought that maybe we could not include the 2 amounts after the comma.
> So if the amount is 1234,56 and I type 1234 it is shown.

That I translated by:
> mathematically, could we translate it as: if I type 1234 I should look for transactions which have amount >= 1234.00 in euros AND amount < 1235.00 in euros so:
> if I have the following amounts in my transactions: 1234.40 1234.51 1234.68 all of these would be returned?
> Negative values as well (-1234.89, -1234.32 ...)? because maybe you don't want to be bothered to type - for outgoing transactions

I think it makes perfect sense from a user's perspective that he would remember an amount in euros but not the details of the cents. Please advise if you think it's not proper?

Implementation is not complex, but the condition is starting to get a little bit complex itself.

Here is an example of results of such a filter locally:
Screenshot 2020-03-03 at 17 46 12

### DIFF
--- a/lib/graphql/transfer.ts
+++ b/lib/graphql/transfer.ts
@@ -261,7 +261,7 @@ export class Transfer extends IterableModel<
   /**
    * Update a standing order
    *
-   * @param transfer  transfer data including at least id and type. For Timed Orders and Sepa Transfers, only category and userSelectedBooking date can be updated 
+   * @param transfer  transfer data including at least id and type. For Timed Orders and Sepa Transfers, only category and userSelectedBooking date can be updated
    * @returns         confirmation id used to confirm the update of Standing order or Transfer for Sepa Transfer / Timed Order
    */
   public async update(transfer: UpdateTransferInput): Promise<ConfirmationRequestOrTransfer> {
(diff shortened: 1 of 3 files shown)


tests/graphql/transaction.spec.ts
```diff
@@ -257,6 +257,28 @@ describe("Transaction", () => {
             name_likeAny: ["1234", "-567", "86.12", "90,1"],
             operator: BaseOperator.Or,
             purpose_likeAny: ["1234", "-567", "86.12", "90,1"],
+            conditions: [
```
hihuz: Existing test cases cover all of the new filter cases, so I didn't add new ones specifically for this.

lib/graphql/transfer.ts
```diff
@@ -261,7 +261,7 @@ export class Transfer extends IterableModel<
   /**
    * Update a standing order
    *
-   * @param transfer  transfer data including at least id and type. For Timed Orders and Sepa Transfers, only category and userSelectedBooking date can be updated 
+   * @param transfer  transfer data including at least id and type. For Timed Orders and Sepa Transfers, only category and userSelectedBooking date can be updated
```
hihuz: Totally unrelated but my editor kept removing this, so I added a separate commit just for this change

PR #86, opened 2020-04-02
Update schema and add splits to TRANSACTION_FIELDS

kolyose:
Add support for transaction splits

Corresponds to https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-app/5045

### DIFF
--- a/lib/graphql/schema.ts
+++ b/lib/graphql/schema.ts
@@ -23,7 +23,7 @@ export type Account = {
   transfer?: Maybe<Transfer>,
   /** Individual tax-related settings per year */
   taxYearSettings: Array<TaxYearSetting>,
-  /**
+  /** 
  * A list of iban/name combinations based on existing user's transactions,
    * provided to assist users when creating new transfers
  **/
@@ -879,6 +879,7 @@ export type Transaction = {
   e2eId?: Maybe<Scalars['String']>,
   mandateNumber?: Maybe<Scalars['String']>,
   fees: Array<TransactionFee>,
+  splits: Array<TransactionSplit>,
   /** The date at which the transaction was booked (created) */
   bookingDate: Scalars['DateTime'],
   directDebitFees: Array<DirectDebitFee>,
@@ -894,7 +895,6 @@ export type Transaction = {
   documentType?: Maybe<DocumentType>,
   foreignCurrency?: Maybe<Scalars['String']>,
   originalAmount?: Maybe<Scalars['Int']>,
-  splits?: Maybe<Array<TransactionSplit>>
 };
 
 export enum TransactionCategory {
@@ -955,15 +955,6 @@ export type TransactionFee = {
   unitAmount?: Maybe<Scalars['Int']>,
   usedAt?: Maybe<Scalars['DateTime']>,
 };
-
-export type TransactionSplit = {
-  __typename?: 'TransactionSplit',
-  id: Scalars['ID'],
-  transactionId:  Scalars['ID'],
-  amount: Scalars['Int'],
-  category: TransactionCategory,
-  userSelectedBookingDate: Scalars['DateTime']
-}
 
 export enum TransactionFeeStatus {
   Created = 'CREATED',
@@ -1048,9 +1039,7 @@ export enum TransactionProjectionType {
   Rebooking = 'REBOOKING',
   CancellationDirectDebit = 'CANCELLATION_DIRECT_DEBIT',
   CancellationSepaCreditTransferReturn = 'CANCELLATION_SEPA_CREDIT_TRANSFER_RETURN',
-  CardTransaction = 'CARD_TRANSACTION',
-  InterestAccrued = 'INTEREST_ACCRUED',
-  CancellationInterestAccrued = 'CANCELLATION_INTEREST_ACCRUED'
+  CardTransaction = 'CARD_TRANSACTION'
 }
 
 export type TransactionsConnection = {
@@ -1063,6 +1052,15 @@ export type TransactionsConnectionEdge = {
    __typename?: 'TransactionsConnectionEdge',
   node: Transaction,
   cursor: Scalars['String'],
+};
+
+export type TransactionSplit = {
+   __typename?: 'TransactionSplit',
+  id: Scalars['ID'],
+  amount: Scalars['Int'],
+  category: TransactionCategory,
+  userSelectedBookingDate: Scalars['DateTime'],
+  transactionId: Scalars['ID'],
 };
 
 export type Transfer = {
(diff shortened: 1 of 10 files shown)


kolyose:
WIP/blocked until we release the corresponding BE change

lib/graphql/schema.ts
```diff
@@ -855,6 +855,7 @@ export type Transaction = {
   e2eId?: Maybe<Scalars['String']>,
   mandateNumber?: Maybe<Scalars['String']>,
   fees: Array<TransactionFee>,
+  splits: Array<TransactionSplit>,
```
francisauwerda, line 858: I'm not sure if those other comments are necessary in this file, but maybe we should leave a comment as to what `splits` is?

lib/graphql/schema.ts
```diff
@@ -1014,9 +1015,7 @@ export enum TransactionProjectionType {
   Rebooking = 'REBOOKING',
   CancellationDirectDebit = 'CANCELLATION_DIRECT_DEBIT',
   CancellationSepaCreditTransferReturn = 'CANCELLATION_SEPA_CREDIT_TRANSFER_RETURN',
-  CardTransaction = 'CARD_TRANSACTION',
-  InterestAccrued = 'INTEREST_ACCRUED',
-  CancellationInterestAccrued = 'CANCELLATION_INTEREST_ACCRUED'
```
francisauwerda, line 1019: Maybe you didn't mean to remove this stuff

hihuz: This was added from us recently, we just merged a huge feature branch to develop along with a few other schema changes (and we didn't update the sdk schema for it yet).

I'm fine with both:
* only including "splits" change in this PR
* doing a bulk schema update with everything missing. In which case doing `schema:generate` against a fresh develop would reveal that these two lines should not be removed, as well as adding a few other overdraft related things to the schema

lib/graphql/schema.ts
hihuz: Necessary or not, they are auto generated from `description` fields in our backend, which adds them as documentation programmatically not only to our schema when you inspect it (for example with GraphQL Playground), but also on our live documentation: https://kontist.dev/docs/#schema-reference 🙂We could argue that some of the existing descriptions are self-explanatory, but we need to keep in mind that external developers, or new kontist developers could still be interested in information that seem redundant to us. 

From a glance it's not clear to me what these "splits" are even if I know the existing backend structure and that they look a lot like existing "transaction meta", so I'd say a bit of description could be useful 🙂 

For instance in the backend taxYearSettingsResolvers, we added a couple of descriptions like "Individual tax-related settings per year" which are now included in the schema, and as a result deployed here in the docs: https://kontist.dev/docs/#taxyearsetting

kolyose: I agree it's a good practice to add some descriptions 👍 However, I'm struggling a bit with making a clear but still short description for the field. I mean, we can't squeeze and put the entire feature with all its details in one field description 🤷‍♂so what do you think guys, smth like this `A list of 2 elements containing metadata for split parts of the transaction` would be meaningful enough? or should I be more verbose/specific? @hihuz @francisauwerda

lib/graphql/schema.ts
kolyose: ok, I will update to latest develop and regenerate the schema

kolyose: unless you want to do it earlier for any reason? @hihuz I mean this PR might be hanging here for a while since @shotexa might need to extend it working on his tickets in the scope of the same `split` feature. We just wanted to release everything split-related to SDK all at once under a single version

lib/graphql/schema.ts
francisauwerda: Cool @hihuz , that's really good to know, thanks for the explanation! I agree, let's add some descriptions to our BE.

I came up with this:
`Information regarding the two separate transactions created when splitting the parent transaction`
But now I'm thinking your suggestion is good too. 😬 Just put anything and we can always improve it if it still looks confusing.

kolyose: I like yours too @francisauwerda let's let @hihuz pick from the options as an independent viewer (not in the context of the feature) lol

hihuz: Hah, so I have to choose 😛 I think there's no need to be too specific about it, what is the reason we are splitting these metadata for? it's to separate by year or something like that right?

lib/graphql/schema.ts
hihuz: We don't really need to do it any time soon, as overdraft will be available only from the native app, which doesn't use typescript for now 🙁so the schema only slightly helps for autocompletion there.

But on the other hand it's our responsiblity to do it, I'll create a PR to add our overdraft Schema changes 👍

lib/graphql/schema.ts
kolyose: you split a transaction into 2 parts - giving each its own amount, category and userSelectedBookingDate - which would lead to more accurate money buckets calculations

hihuz: ok, I guess "Metadata resulting from splitting a transaction" makes sense to me 🤔 not sure if proper english 😄 Otherwise let's go with Francis' suggestion 🙂

kolyose: ok, so let @francisauwerda decide now 😆 kidding, I compiled both your options to `Metadata of separate pseudo-transactions created when splitting the parent transaction`

lib/graphql/schema.ts
hihuz: Added overdraft changes in this PR: https://github.com/kontist/js-sdk/pull/87 🙂

lib/graphql/schema.ts
```diff
@@ -1055,6 +1054,14 @@ export type TransactionsConnectionEdge = {
   cursor: Scalars['String'],
 };
 
+export type TransactionSplit = {
+   __typename?: 'TransactionSplit',
+  id: Scalars['ID'],
```
francisauwerda, line 1059: should this be optional? `id?: Scalars['ID']` since when creating a transaction split there won't be an ID created yet?

lib/graphql/schema.ts
```diff
@@ -1055,6 +1054,14 @@ export type TransactionsConnectionEdge = {
...
+  amount: Scalars['Int'],
+  category: TransactionCategory,
+  userSelectedBookingDate?: Maybe<Scalars['DateTime']>,
```
francisauwerda, line 1062: Should this be mandatory? And we pass `null` when there is no booking date provided?
So like this
```
userSelectedBookingDate: Maybe<Scalars['DateTime']>,`
```
I thought that's what the `Maybe` meant. 
So basically I'm saying, can we/should we remove the question mark?

shotexa: `export type Maybe<T> = T | null;` as defined in the top of the file
when we create a split, userSelectedBookingDate is optional, Maybe allowed 'null' but '?' allowed 'undefined', so it can either be `null`, 'undefined' (so we just don't pass anything in a query) or 'Date'

PR #88, opened 2020-04-06
Add  model with a method to update subscription plan

dmytroyarmak:


lib/graphql/subscription.ts
```diff
@@ -0,0 +1,34 @@
...
+export class Subscription {
+  constructor(protected client: GraphQLClient) {}
+
+  /**
+   * Updates subscription plan
+   *
+   * @param args  query parameters including new plan name
+   * @returns     result of update
+   */
+  public async updateSubscriptionPlan(
```
hihuz, line 28: what about `updatePlan` or just `update` instead of `updateSubscriptionPlan`?

For most existing methods we don't repeat the model name (e.g. `transaction.categorize()`)

dmytroyarmak: agree, fixed in 8e3a79d

lib/graphql/schema.ts
```diff
@@ -1055,6 +1072,14 @@ export type TransactionsConnectionEdge = {
   cursor: Scalars['String'],
 };
 
+export type TransactionSplit = {
```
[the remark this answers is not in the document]
dmytroyarmak: done in 7198315

tests/graphql/subscription.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { expect } from "chai";
+import * as sinon from "sinon";
+
+import { Client } from "../../lib";
+import { Subscription } from "../../lib/graphql/subscription";
+import { PurchaseType, UpdateSubscriptionPlanResult } from "../../lib/graphql/schema";
+
+
```
goodfellamilos, line 8: why 2 spaces? 😄

tests/graphql/subscription.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+describe("Subscription", () => {
+  let sandbox: sinon.SinonSandbox;
+  let client: Client;
+
+  beforeEach(async () => {
```
goodfellamilos, line 13: no need for `async`

PR #89, opened 2020-04-20
Expose flow types

hihuz:
Let's expose flow types for our schema. It will be useful at least for our native app.

Thanks @johannespfeiffer for the suggestion to use graphql-codegen directly for this. Made it quite simple.

I'm not sure of the best approach to expose them though. So I have simply put them in the build next to the `schema.d.ts`.

It forces me to import them as "kontist/dist/lib/graphql/schema.flow" in the native app, but I'm doing so only once. I think this is OK, and I'm not sure there is a better way, considering our lib is written in typescript and not flow.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const CopyPlugin = require('copy-webpack-plugin');
 
 module.exports = {
   entry: './dist/lib/index.js',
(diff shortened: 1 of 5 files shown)


webpack.config.js
```diff
@@ -10,4 +11,9 @@ module.exports = {
     path: path.resolve(__dirname, 'dist'),
     library: 'Kontist'
   },
+  plugins: [
+    new CopyPlugin([
+      { from: './lib/graphql/schema.flow.js', to: './lib/graphql/schema.flow.js'}
```
hihuz, line 16: our build steps are `tsc && webpack` and typescript compiler can't handle this "flow" typed file, even if I `allowJs: true`. So I think the best approach is simply to copy it as is in the build folder with webpack

PR #90, opened 2020-04-20
Update vulnerable dependencies

hihuz:
Unfortunately my publish failed due to `npm audit` finding some vulnerabilities: https://github.com/kontist/js-sdk/runs/602299503

I am fixing them in this PR. 

I ran schema:generate after the update to double check, and in the diff, only changes were `,` -> `;` in types, so I'd say it's safe.


hihuz:
Ah, yeah, I remember reading about this issue. Looks like they fixed it indeed!

PR #91, opened 2020-04-22
Add helper account stats 6852

amodesKontist:
Addresses
https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-backend/6852

Makes `account.getAccountStats()` available in the sdk.

Please make sure to have this BE running:
https://github.com/kontist/figure-backend/pull/6845

### DIFF
--- a/lib/graphql/account.ts
+++ b/lib/graphql/account.ts
@@ -55,7 +55,7 @@ export class Account extends Model<AccountModel> {
    *
    * @returns current user account statistics
    */
-  public async getAccountStats(): Promise<AccountStats | null> {
+  public async getStats(): Promise<AccountStats | null> {
     const result: Query = await this.client.rawQuery(GET_ACCOUNT_STATS);
     return result.viewer?.mainAccount?.accountStats ?? null;
   }
--- a/lib/graphql/account.ts
+++ b/lib/graphql/account.ts
@@ -1,7 +1,7 @@
 import { KontistSDKError } from "../errors";
 import { Model } from "./model";
 import { ResultPage } from "./resultPage";
-import { Account as AccountModel, Query } from "./schema";
+import { Account as AccountModel, Query, AccountStats } from "./schema";
 
 const GET_ACCOUNT = `query {
   viewer {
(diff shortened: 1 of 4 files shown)


lib/graphql/account.ts
```diff
@@ -13,10 +13,31 @@ const GET_ACCOUNT = `query {
     }
   }
 }`;
+const GET_ACCOUNT_STATS = `query {
+  viewer {
+    mainAccount {
+      accountStats {
+        accountBalance
+				main
```
amodesKontist, line 21: In VS code this doesnt look that weird 🤔

lib/graphql/account.ts
```diff
@@ -28,4 +49,14 @@ export class Account extends Model<AccountModel> {
     const result: Query = await this.client.rawQuery(GET_ACCOUNT);
     return result.viewer?.mainAccount || null;
   }
+
+  /**
+   * Different statistics to the users account balance e.g. yours, unknow, vatTotal, taxMissing ...
+   *
+   * @returns current user account statistics
+   */
+  public async getAccountStats(): Promise<AccountStats | null> {
```
amodesKontist, line 58: was not sure how to name it.
now its used like `account.getAccountStats()` maybe just `account.getStats()` is better? 🤔

hihuz: I think `getStats` would be better for our SDK 🙂 We usually don't repeat model name in methods, e.g. `transaction.categorize`, `transfer.createOne`... there are exceptions but let's try to keep consistency higher by using `getStats` name 👍

amodesKontist, line 58: Also I was thinking of making it configurable by passing an array of strings with fields you really want to have,
e.g `getAccountStats(['accountBalance','vatTotal','taxTotal'])`. And if no parameter is provided, return all fields like it is at the moment. But not sure if we should keep it simple. What do you think about it  🤔

hihuz: This is an idea, but at that point, consumers would be probably better off writing the query on their own with rawQuery directly.

For now we don't provide such possibility of filtering in any other helper. I'd say we should keep it that way for now. 

We will always have the possibility later on to enrich these helpers, e.g. `getStats({ fields: [...] })` or so.

lib/graphql/account.ts
```diff
@@ -28,4 +49,14 @@ export class Account extends Model<AccountModel> {
     const result: Query = await this.client.rawQuery(GET_ACCOUNT);
     return result.viewer?.mainAccount || null;
   }
+
+  /**
+   * Different statistics to the users account balance e.g. yours, unknow, vatTotal, taxMissing ...
```
hihuz, line 54: unknow -> unknown 🙂

lib/graphql/account.ts
hihuz: these are tabs instead of spaces 🙂

amodesKontist:
> Looks great 👍
> 
> Let's fix these 2 small issues and agree on backend naming before merging it 🙂

Pushed just `stats` instead of `accountStats`, like in BE :)

lib/graphql/account.ts
```diff
@@ -13,10 +13,31 @@ const GET_ACCOUNT = `query {
     }
   }
 }`;
+const GET_ACCOUNT_STATS = `query {
+  viewer {
+    mainAccount {
+      accountStats {
```
hihuz, line 19: so it should now be `stats` here too, right? 😛

amodesKontist: dont check the changes too fast! 😅  You dont give me the chance to hide my mistakes  😜

PR #92, opened 2020-04-23
Add helpers for transaction splits

pstockschlaeder:
**Addresses [figure-backend/6752](https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-backend/6752)**

Adds helpers to create, update and delete [transaction splits](https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-app/4952).

After the changes have been released, we should also update the [documentation](https://kontist.dev/sdk/#categorize-a-transaction) in the [kontist.github.io](https://github.com/kontist/kontist.github.io) repository.


tests/graphql/transaction.spec.ts
```diff
@@ -204,6 +209,146 @@ describe("Transaction", () => {
...
+        },
+        {
+          id: 2,
+          amount: -500,
+          category: TransactionCategory.Vat_19
+        },
+      ];
+      const transactionData = createTransaction({
+        amount: -1000,
+        category: TransactionCategory.Private,
```
francisauwerda, line 241: I don't think this is possible, we won't be able to have a transaction with a category as "PRIVATE" and have splits. It doesn't make a difference for the tests, it just might be unnecessary. But anyway, the tests aren't testing that part, so the tests are good!

pstockschlaeder: Thanks for the input 🙌! You are right, after a split the parent transaction's `category` will be cleared. I updated the tests in https://github.com/kontist/js-sdk/pull/92/commits/8cdd542e662ca14940bdf50393a80b936e24eb07 to now consist of real-world examples. Also, I added some verbosity which makes the tests a bit lengthy but hopefully more comprehensible.

francisauwerda:
Changelog changes look good!

PR #93, opened 2020-04-23
Fix some misc typos

pstockschlaeder:
This PR fixes some small typos we happened to find.

Please note that this is a **breaking change** (rename of `StandingOrderReoccurenceType` to `StandingOrderReoccurrenceType`).


lib/graphql/schema.flow.js
```diff
@@ -297,7 +297,7 @@ export type CreateTransferInput = {|
   /** The end to end ID of the transfer */
   e2eId?: ?$ElementType<Scalars, 'String'>,
   /** The reoccurrence type of the payments for Standing Orders */
-  reoccurrence?: ?StandingOrderReoccurenceType,
+  reoccurrence?: ?StandingOrderReoccurrenceType,
```
hihuz, line 300: Ah 🙁good catch, but that sucks.

Unfortunately fixing this one is going to be a bit more involved.
These schemas are auto generated (with `npm run schema:generate`) from our backend types, which means the typo would reappear every time we generate it again.

Fixing it on the backend first would do it, but it is also a breaking change.

I think it is acceptable (considering the typo is in the enum name and not the values, we can safely update the backend), but any consumer updating the sdk with this change will have to update this type as well in its own code (and both our webapp and native app use it).
So we should at least document it in the `CHANGELOG.md`

Shall we add under `## [Unreleased]`:
```
### Changed
- *Breaking*: renamed StandingOrderReoccurenceType to StandingOrderReoccurrenceType
```
or something similar?

In the meantime I'll create the corresponding backend PR

pstockschlaeder: Thanks for paying attention 🙏. I intentionally put it in a separate commit but forgot to mention it explicitly in the PR.

Btw: What's your suggestion on such changes in general? One can even argue argue that it's not worth a breaking change. One example that comes to my mind is the HTTP `referer` header which still is missing an `r` and most likely will always be 😏.

hihuz: Heh 🙂

Yeah, I'm unsure about it in this case I hesitated before commenting, I think each case can be considered individually rather than having a general rule. 🤔 

On the one hand it's not that bad to have a small typo here, but on the other hand fixing it is still straightforward and safe (by that I mean that there is no risk of breakage for users if they don't update the lib), as long as we document it. 

Let's get maybe an other opinion on it? cc @johannespfeiffer 

I'd honestly be fine with both approaches

pstockschlaeder: > I'd honestly be fine with both approaches

Me too 🙌. We can also revert https://github.com/kontist/js-sdk/pull/93/commits/5261833a80ad7a098ff7ce570fed386e474c1ea4 and merge only the non-breaking changes.

johannespfeiffer: Let's do it. We do not have that many users as HTTP (yet) and the change for the users of the SDK will be only very minor.

tests/helpers.ts
```diff
@@ -29,7 +29,7 @@ export const createTransaction = (
     bookingDate: new Date(),
     directDebitFees: [],
     fees: [],
-    id: Math.random.toString(),
+    id: Math.random().toString(),
```
johannespfeiffer, line 32: Good catch.

lib/graphql/schema.flow.js
pstockschlaeder: Corresponding backend PR: https://github.com/kontist/figure-backend/pull/6870

PR #95, opened 2020-05-04
Fetch, add, delete google pay card tokens

p-janik:
Allow to fetch, add, and delete google pay card token reference ids for given wallet id. 
Part of https://github.com/kontist/figure-backend/pull/6952

### DIFF
--- a/lib/graphql/card.ts
+++ b/lib/graphql/card.ts
@@ -10,6 +10,9 @@ import {
   MutationReplaceCardArgs,
   MutationUpdateCardSettingsArgs,
   Query,
+  GooglePayCardToken,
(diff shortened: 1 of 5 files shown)


lib/graphql/card.ts
```diff
@@ -214,6 +221,36 @@ export const SET_CARD_HOLDER_REPRESENTATION = `mutation setCardHolderRepresentat
   )
 }`;
 
+export const ADD_GOOGLE_PAY_CARD_TOKEN = `mutation(
+  $id: String!,
```
hihuz, line 225: Shall we make it consistent by having `,` for both lines or none of them (and in the delete mutation too)?

Yes I know we have this inconsistency everywhere 🙁

PR #96, opened 2020-05-06
Update schema with added user fields

hihuz:
Let's update schema with latest added user fields.


lib/graphql/schema.flow.js
```diff
@@ -454,12 +525,18 @@ export type Mutation = {|
   categorizeTransaction: Transaction,
   /** Create Overdraft Application  - only available for Kontist Application */
   requestOverdraft?: ?Overdraft,
+  /** Activate Overdraft Application  - only available for Kontist Application */
+  activateOverdraft?: ?Overdraft,
```
lera, line 529: oh! thanks! forgot about it🤦‍♀️

PR #100, opened 2020-05-29
Remove redundant taxCutoffLine support

kolyose:
In order to remove the property from BE we have to remove it from the SDK first.

Corresponds to a ticket https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-app/3581#issuecomment-629274380

### DIFF
--- a/lib/graphql/schema.flow.js
+++ b/lib/graphql/schema.flow.js
@@ -14,7 +14,7 @@ export type Scalars = {|
 
 /** The bank account of the current user */
 export type Account = {|
-   __typename?: 'Account',
+  __typename?: 'Account',
   iban: $ElementType<Scalars, 'String'>,
   cardHolderRepresentation?: ?$ElementType<Scalars, 'String'>,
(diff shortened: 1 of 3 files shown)


kolyose:
yeah, I just noticed this 🤦 removing it now

kolyose:
Reverted back the changes in the schemas since they are going to be autogenerated back anyways

lib/graphql/schema.flow.js
```diff
@@ -14,7 +14,7 @@ export type Scalars = {|
 
 /** The bank account of the current user */
 export type Account = {|
-   __typename?: 'Account',
```
hihuz, line 17: We still have a number of apparently unrelated changes here 🤔 But if you had them when running the `schema:generate` command, all good

kolyose: yes, that's what it generated 🤷‍♂️

PR #101, opened 2020-06-03
Update schema with personal note

Campalo:
[Ticket 4831](https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-app/4831) from the Transaction note epic. 

Update the schema with the `personalNote` field and the `updateTransaction` method. 
Deprecate the `categorizeTransaction` method which is now included in the `updateTransaction` method. 
Extend `transfer` and `transaction` tests.

### DIFF
--- a/lib/graphql/transaction.ts
+++ b/lib/graphql/transaction.ts
@@ -228,7 +228,7 @@ export class Transaction extends IterableModel<TransactionModel> {
    * @param args   query parameters including category, userSelectedBookingDate and personalNote
    * @returns      the transaction with updated categorization data and updated personalNote
    */
-  public async updateMeta(args: MutationUpdateTransactionArgs) {
+  public async update(args: MutationUpdateTransactionArgs) {
     const result = await this.client.rawQuery(UPDATE_TRANSACTION, args);
     return result.updateTransaction;
   }
--- a/lib/graphql/transaction.ts
+++ b/lib/graphql/transaction.ts
@@ -4,6 +4,7 @@ import {
   AccountTransactionsArgs,
   BaseOperator,
   MutationCategorizeTransactionArgs,
+  MutationUpdateTransactionArgs,
   MutationCreateTransactionSplitsArgs,
   MutationDeleteTransactionSplitsArgs,
   MutationUpdateTransactionSplitsArgs,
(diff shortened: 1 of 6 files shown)


tests/graphql/transaction.spec.ts
```diff
@@ -173,7 +173,7 @@ describe("Transaction", () => {
     });
   });
 
-  describe("#categorize", () => {
+  describe("#update", () => {
```
lera, line 176: shouldn't we leave old tests? because method will still be used for a while

lib/graphql/transaction.ts
```diff
@@ -204,6 +222,17 @@ export class Transaction extends IterableModel<TransactionModel> {
     return result.categorizeTransaction;
   }
 
+  /**
+   * Updates a transaction
+   *
+   * @param args   query parameters including category, userSelectedBookingDate and personalNote
+   * @returns      the transaction with updated categorization data and updated personalNote
+   */
+  public async updateMeta(args: MutationUpdateTransactionArgs) {
```
lera, line 231: shouldn't it be called updateTransaction? what do you think?

hihuz: I would think just `update`, because `updateTransaction` would be used as `models.transaction.updateTransaction(...)` which makes "Transaction" redundant.

We also have the same approach in general for model methods, e.g. `models.transaction.categorize`, `models.transfer.createOne`, `models.transfer.update`... It's not always consistent but recently we have tried to stick to the convention of not repeating model name in method name.

Between `transaction.updateMeta` and `transaction.update` I would also have a slight preference for `transaction.update` (because it would allow more generic updates than just meta, if we have the requirement later), but I wouldn't object in other name.

Campalo: Thanks for sharing your thoughts about the new method name. I'll go for `update()` as this follows the new convention of  "not repeating model name in the method name", so we stay consistent.

tests/graphql/transaction.spec.ts
Campalo: Oups right. I'll put them back.

PR #103, opened 2020-06-09
Update Schema and taxPastYearsAmount spelling

francisauwerda:
Mainly for the spelling mistake of taxPastYearsAmount

### DIFF
--- a/lib/graphql/schema.flow.js
+++ b/lib/graphql/schema.flow.js
@@ -14,7 +14,7 @@ export type Scalars = {|
 
 /** The bank account of the current user */
 export type Account = {|
-  __typename?: 'Account',
+   __typename?: 'Account',
   iban: $ElementType<Scalars, 'String'>,
   cardHolderRepresentation?: ?$ElementType<Scalars, 'String'>,
   balance: $ElementType<Scalars, 'Int'>,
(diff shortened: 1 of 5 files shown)


lib/graphql/schema.flow.js
```diff
@@ -14,7 +14,7 @@ export type Scalars = {|
 
 /** The bank account of the current user */
 export type Account = {|
-  __typename?: 'Account',
+   __typename?: 'Account',
```
francisauwerda, line 17: Do I have some incorrect setting that formatted this with an extra space??

p-janik: I had the same yesterday. Decided to include my schema changes :)

hihuz: I also have the same locally, the change was introduced by Sergey when he ran the command previously, see: https://github.com/kontist/js-sdk/pull/100#discussion_r432493232

It might have to do with environment settings or something, I'd rather keep the version that most of us have, so I'm fine with keeping these extra changes

lib/graphql/schema.flow.js
```diff
@@ -113,20 +113,26 @@ export type AccountStats = {|
   taxTotal: $ElementType<Scalars, 'Int'>,
   /** The amount of tax that is owed in the current year */
   taxCurrentYearAmount: $ElementType<Scalars, 'Int'>,
-  /** The amount of tax that was owed last year */
-  taxPastYearAmount?: ?$ElementType<Scalars, 'Int'>,
+  /** The amount of tax that was owed for all past years combined */
+  taxPastYearsAmount?: ?$ElementType<Scalars, 'Int'>,
```
francisauwerda, line 117: Here are my only changes.

PR #104, opened 2020-06-09
Bump version 0.28.7

francisauwerda:


francisauwerda:
Is that correct? Not that there are 2 commits with the same description, but that one updates the `package.json` and the other updates `package-lock.json` with the same version bump ?

francisauwerda:
I'm going to start again

PR #105, opened 2020-06-09
Expose predicted* meta fields in categorizeTransaction response

p-janik:
Respond with predicted category on categorize transaction response.

This mutation will be deprecated soon. I see that I should use extended https://github.com/kontist/js-sdk/pull/101 later.

Since predictedCategory will be used for now just for popup feature in mobile app, I decided to not include it in every transaction response.


p-janik:
This won't be needed. We can fetch predicted* details with transaction details query (which currently is rawQuery on mobile)

PR #109, opened 2020-06-17
Update User type with couponCodeOffer field

goodfellamilos:
New field added to `User` in [figure-backend/7346](https://github.com/kontist/figure-backend/pull/7346).

Zenhub: [#5344](https://app.zenhub.com/workspaces/kontist-engineering-5be06ab44b5806bc2bf14eca/issues/kontist/figure-app/5344)


goodfellamilos:
Yes, I see what you mean. At the moment there is no need for this field to be returned as a part of user query. Do you think I should still add it now? Or maybe later when there is a need for it?

johannespfeiffer:
> Do you think I should still add it now? Or maybe later when there is a need for it?
Later is fine if we do not need it now.

PR #111, opened 2020-06-22
Extend schema with categorizationScreenShown field

p-janik:


lib/graphql/schema.flow.js
```diff
@@ -1683,7 +1683,6 @@ export type User = {|
   /** The current state of user's Kontist account based on his subscription plan */
   accountState?: ?AccountState,
   businessTradingName?: ?$ElementType<Scalars, 'String'>,
-  /** Coupon code assigned to the user that can be redeemed during subscription update */
```
lera, line 1686: why is it removed?

hihuz: I think it was added manually in previous PR. comment does not exist on backend, so `schema:generate` would remove it.

I'll ping Comrades squad to see if they want to add it to backend.

hihuz: @p-janik the missing description was added to backend, do you mind running the `schema:generate` again so that we don't remove it?

goodfellamilos:
@p-janik Should this one be merged?

PR #117, opened 2020-07-08
Update dependencies

johannespfeiffer:
Just updating everything to the most recent version. Skipped TypeScript because of problems mocking a ES6 import dependency.


johannespfeiffer:
After the updated packages it became obvious that we missed to explicitly add a polyfill for fetch on nodejs. This is done via a "ponyfill" of cross-fetch now.

PR #125, opened 2020-07-16
Update schema for self employed field

hihuz:
I am adding `isSelfEmployed` field to schema. 

The large number of changes in the schema was caused by a recent update of `graphql-codegen` to 1.16.3.
It looks like they introduced some changes in this version reordering the entities in the schema.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4061,8 +4061,7 @@
     "graphql": {
       "version": "15.3.0",
       "resolved": "https://registry.npmjs.org/graphql/-/graphql-15.3.0.tgz",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -4061,8 +4061,7 @@
     "graphql": {
       "version": "15.3.0",
       "resolved": "https://registry.npmjs.org/graphql/-/graphql-15.3.0.tgz",
-      "integrity": "sha512-GTCJtzJmkFLWRfFJuoo9RWWa/FfamUHgiFosxi/X1Ani4AVWbeyBenZTNX6dM+7WSbbFfTo/25eh0LLkwHMw2w==",
-      "dev": true
```
hihuz, line 4065: This was changed in a previous PR (devDependencies -> dependencies) but we didn't update package-lock for it

PR #128, opened 2020-07-28
Update schema

amodesKontist:
Add the gender to the `UserUpdateInput`.


lib/graphql/schema.flow.js
```diff
@@ -1858,6 +1858,7 @@ export type UserUpdateInput = {|
   vatNumber?: ?$ElementType<Scalars, 'String'>,
   vatRate?: ?$ElementType<Scalars, 'Int'>,
   language?: ?$ElementType<Scalars, 'String'>,
+  gender?: ?$ElementType<Scalars, 'String'>,
```
lera, line 1861: shouldn't it be an enum?

amodesKontist: updated it :)

PR #145, opened 2020-09-02
Build correct verifier

johannespfeiffer:
This PR addresses an issue where a user could not log in.
Reason was an incorrect implementation of the https://tools.ietf.org/html/rfc7636#appendix-A algorithm.

To be specific - the log in failed every time when the code challenge contained more than one character of "+" or "/".

`replace` is not `replaceAll`


johannespfeiffer:
Fix requires corresponding backend pr.

hihuz (approved):
Wow.

Thanks a lot 👍

PR #150, opened 2020-10-26
Update schema with refund dd and review flow types

amodesKontist:
Updates the schema because we would like to use review flow types
in https://github.com/kontist/webapp/pull/1988
and https://github.com/kontist/app/pull/6165


Campalo (approved):
Works well with the local https://github.com/kontist/webapp/pull/1988
Better types too. LGTM 👍

PR #152, opened 2020-11-27
Fix or skip some unit tests

haipham-kontist:
Currently, 4 unit tests are failing. I'm fixing 2 of them.

There seems to be something wrong with `when auth token is present` test suite, the stub functions don't get called as expected.

```js
expect(subscriptionClientStub.callCount).to.equal(1);
```

This returns `0` instead.

@johannespfeiffer @lera please let me know what you think.


haipham-kontist:
I see,
- if we do `npm install` then `yarn test`, all the tests will pass
- if we do `yarn` then `yarn test`, some of the tests are failed. I think it's because of deprecated dependencies.

I'll close the ticket for now.

PR #153, opened 2020-11-30
Allow making GQL request to public endpoint

haipham-kontist:
Allow making GQL request to public endpoint.


README.md
```diff
@@ -71,6 +71,17 @@ Optionally, this method accepts a number as an argument to specify after how man
...
 ```
 
+**Note:** to create a client that can handle public request, you need to initialize `Client` without `clientId`.
+
+```typescript
+const client = new Client({
+  redirectUri: REDIRECT_URI,
+  scopes: ["transactions"],
+  clientSecret,
+  state
```
johannespfeiffer, line 81: Without authorization it does not make sense to give the `redirectUri`, `clientSecret` params etc.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kontist",
-  "version": "0.34.7",
+  "version": "0.34.8-2",
```
johannespfeiffer, line 3: Please remove before merging into main branch.

tests/helpers.ts
```diff
@@ -11,12 +11,23 @@ export const redirectUri = "https://localhost:3000/auth/callback";
 export const scopes = ["transactions"];
 export const state = "25843739712322056";
 
-export const createClient = (opts = {}) => {
+type ClientOpts = {
+  baseUrl?: string;
+  baseSubscriptionUrl?: string;
+  clientId?: string | undefined;
```
johannespfeiffer, line 17: Do we need `| undefined` here or is this already done by `?`?

tests/helpers.ts
```diff
@@ -11,12 +11,23 @@ export const redirectUri = "https://localhost:3000/auth/callback";
 export const scopes = ["transactions"];
 export const state = "25843739712322056";
 
-export const createClient = (opts = {}) => {
+type ClientOpts = {
```
johannespfeiffer, line 14: Instead of duplicating the type here, can we use the one from `lib/types.ts`?
We could make it optional by using `Partial<ClientOpts>`

tests/helpers.ts
```diff
@@ -11,12 +11,23 @@ export const redirectUri = "https://localhost:3000/auth/callback";
...
+  oauthClient?: any;
+  verifier?: string;
+  state?: string;
+};
+
+export const createClient = (opts: ClientOpts = {}) => {
   return new Client({
-    clientId,
+    // opts.clientId can be undefined
+    clientId: opts.hasOwnProperty('clientId') ? opts.clientId : clientId,
```
johannespfeiffer, line 27: nit:
```suggestion
    clientId: opts.clientId ?? clientId,
```

.nycrc
```diff
@@ -2,9 +2,9 @@
     "cache": false,
     "check-coverage": true,
     "branches": 93,
-    "lines": 98,
-    "functions": 98,
-    "statements": 98,
+    "lines": 99,
+    "functions": 99,
+    "statements": 99,
```
johannespfeiffer, line 7: 👍

johannespfeiffer (approved):
Looks great!
Could you add a small note to the `CHANGELOG.md`? Beside that only minor questions.

lib/graphql/client.ts
```diff
@@ -42,15 +42,17 @@ export class GraphQLClient {
       [key: string]: any;
     },
   ): Promise<RawQueryResponse> => {
-    if (!this.auth.tokenManager.token) {
-      throw new UserUnauthorizedError();
+    if (this.auth) {
```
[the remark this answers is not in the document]
haipham-kontist: hi @zlu-kontist , this PR is for making `auth` optional, hence to make request to public endpoint.

so if `!this.auth`, it will not include authorization header in the request.

lib/client.ts
```diff
@@ -4,17 +4,17 @@ import { getModels } from "./graphql";
...
 
-export class Client {
+export class Client<T extends Partial<ClientOpts> = ClientOpts> {
   constructor(
-    opts: ClientOpts,
+    opts: T,
     baseUrl = opts.baseUrl || KONTIST_API_BASE_URL,
     baseSubscriptionUrl = opts.baseSubscriptionUrl || KONTIST_SUBSCRIPTION_API_BASE_URL,
-    public auth = new Auth(baseUrl, opts),
+    public auth: T extends { clientId?: string | undefined } ? Auth : undefined = opts.clientId ? new Auth(baseUrl, opts as ClientOpts) : undefined as any,
```
haipham-kontist, line 12: `opts.clientId` can be `""`. I'll update the code to check for it

johannespfeiffer:
```
    public auth: T extends { clientId?: string | undefined } ? Auth : undefined ...
```
We want to test if `T` extends the type that *has* this `clientId`. So it should be
```
    public auth: T extends { clientId: string } ? Auth : undefined ...
```

johannespfeiffer: So what this means is:
`auth` should be of type `Auth` if the type of `opts` (which is `T`) contains the `clientId`. If that is not the case the type of `auth` should be `undefined`.

haipham-kontist: I understand that part, but doing so will result in eslint error `Object is possibly 'undefined'.`

For example
```
tests/graphql/client.spec.ts:22:7 - error TS2532: Object is possibly 'undefined'.

22       client.auth.tokenManager.setToken("dummy-access-token");
         ~~~~~~~~~~~
````

Comment (haipham-kontist) [reply lib/client.ts]: The issue is because of the check `opts.clientId ?`, it would return false if `opts.clientId` is empty string.

File: lib/graphql/client.ts (hunk 7 above)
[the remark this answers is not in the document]
Comment (zlu-kontist) [reply lib/graphql/client.ts:45]: I guess I'm a bit confused.  if this.auth doesn't exsit, why should we include an auth header?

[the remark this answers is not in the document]
Comment (haipham-kontist) [reply lib/graphql/client.ts:45]: I mean if `this.auth` does not exist, we wouldn't include an auth header.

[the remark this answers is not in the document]
Comment (zlu-kontist) [reply lib/graphql/client.ts:45]: Actually, would it make sense to leverage factory pattern here by scoping requests without auth to only public apis and require auth for all private apis.

File: lib/client.ts (hunk 8 above)
Comment (haipham-kontist) [reply lib/client.ts]: @johannespfeiffer can you help to double check please? Thanks.

File: lib/graphql/client.ts (hunk 7 above)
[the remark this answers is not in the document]
Comment (haipham-kontist) [reply lib/graphql/client.ts:45]: actually, I just implement from what @johannespfeiffer suggested 😅

@johannespfeiffer wdyt?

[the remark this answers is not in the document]
Comment (johannespfeiffer) [reply lib/graphql/client.ts:45]: Scoping requests without auth is not easy, as the schema is autogenerated from GraphQL.

File: lib/graphql/client.ts [hunk 9]
```diff
@@ -141,14 +150,16 @@ export class GraphQLClient {
...
+      try {
+        await this.auth.tokenManager.refresh();
+      } catch (error) {
+        Object.values(this.subscriptions).forEach(({ onError }) => {
+          if (typeof onError === "function") {
+            onError(error);
+          }
+        });
+      }
     }
```
johannespfeiffer, line 163: we could save one if here
```suggestion
    try {
      await this.auth?.tokenManager.refresh();
    } catch (error) {
      Object.values(this.subscriptions).forEach(({ onError }) => {
        if (typeof onError === "function") {
          onError(error);
        }
      });
    }
```

haipham-kontist: I think it would be more readable with the if 🤔

lib/graphql/client.ts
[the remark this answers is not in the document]
johannespfeiffer: If we want to reduce code duplication and number of `if`s we could use a helper method like this. (not sure if it is still easy enough to read)

```
  private getAuthHeader = (): [string, string] | false => {
    if (!this.auth) {
      return false;
    }

    if (!this.auth.tokenManager.token) {
      throw new UserUnauthorizedError();
    }

    return ["Authorization", `Bearer ${this.auth.tokenManager.token.accessToken}`];
  }
```

and then use it with
```
    const auth = this.getAuthHeader();
    auth && this.client.setHeader(...auth);
```

respective
```
    const auth = this.getAuthHeader();
    const connectionParams = auth ? { [auth[0]]: auth[1] } : {};
```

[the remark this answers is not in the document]
haipham-kontist: very nice, thanks!

PR #162, opened 2021-02-19
Update schema

amodesKontist:


lib/graphql/schema.flow.js
```diff
@@ -1,4 +1,4 @@
-/* @flow */
+// @flow
```
Campalo, line 1: Why changing the comment syntax ? Both are fine, just wondering :)

Campalo (approved):
👍

lib/graphql/schema.flow.js
```diff
@@ -30,7 +30,6 @@ export type User = {|
   __typename?: 'User',
   email: $ElementType<Scalars, 'String'>,
   createdAt: $ElementType<Scalars, 'DateTime'>,
-  vatCutoffLine?: ?$ElementType<Scalars, 'DateTime'>,
```
Campalo, line 33: I saw that this deleted line is linked to your previous prs from BE and webapp 👍

lib/graphql/schema.flow.js
amodesKontist: I just ran `yarn schema:generate`  🤔

lib/graphql/schema.flow.js
amodesKontist: yes its the reason why I do this update.
Problem is that there is a test on BE that uses the sdk to request `vatCutoffline`.
So the test from this PR always fails https://github.com/kontist/backend/pull/9220 so I update the sdk first. 
Kind of circular depenedency between the test and sdk 😅

PR #167, opened 2021-02-22
Re-generate schema for split's categorizationType

kolyose:
Requests for split's `categorizationType` attribute and updates the schemas accordingly


p-janik:
@kolyose do you know why we such diff for `lib/graphql/schema.flow.js`?

amodesKontist:
I think Sylvain once told me it is like this because when you use npm instead of yarn or have different yarn/npm versions 🤔

kolyose:
@p-janik I can see the majority of the changes are whitespaces 🤔

kolyose:
so I tried updating `yarn` to the latest version(s) (`1.22.10` & `2.4.0`), also I tried it with `npm` - all of them generate me the same schema with all the whitespaces removed at the eol 🤷‍♂️

amodesKontist:
> so I tried updating `yarn` to the latest version(s) (`1.22.10` & `2.4.0`), also I tried it with `npm` - all of them generate me the same schema with all the trailing whitespaces removed 🤷‍♂️

Ok I checked with yarn v.1.22.10 and npm, and I get the trailing whitespaces :man-shrugging: 
Maybe we can create housekeeping ticket to investigate more

amodesKontist (approved):

PR #178, opened 2021-03-11
Update GraphQL schema with changes to signPOA mutation

dmytroyarmak:


fabribet (approved):
It seems that we are missing the `UserDependentType` type 🤔

dmytroyarmak:
> It seems that we are missing the `UserDependentType` type 🤔

you are right, it looks like enum is not registered for GraphQl

PR #189, opened 2021-04-09
Update schema

amodesKontist:
Update schema with invoicing dashboard types


lib/graphql/schema.flow.js
```diff
@@ -163,316 +170,316 @@ export const TaxPaymentFrequencyValues = Object.freeze({
...
-  Vn: 'VN',
-  Vu: 'VU',
-  Wf: 'WF',
-  Ws: 'WS',
-  Xk: 'XK',
-  Ye: 'YE',
-  Yt: 'YT',
-  Za: 'ZA',
-  Zm: 'ZM',
+  De: 'DE',
```
Campalo, line 204: Is it possible to save the file without this extra space at the end of the lines? makes it harder to read

amodesKontist: We didn't figured it out yet why for some devs there is this space and for some not. You can just check the changes in the ts schema, as flow is always the same

lib/graphql/schema.ts
```diff
@@ -161,6 +163,11 @@ export type UserInvoiceArgs = {
 };
 
 
+export type UserInvoicingDashboardDataArgs = {
+  pageNumber: Scalars['Float'];
```
Campalo, line 167: Should it rather be `INTEGER` ? I mean a page will always be a whole number, there will never be decimal digits.

lib/graphql/schema.ts
```diff
@@ -1355,13 +1332,44 @@ export type Invoice = {
 
 export type InvoiceProductOutput = {
   __typename?: 'InvoiceProductOutput';
-  id: Scalars['String'];
+  id?: Maybe<Scalars['String']>;
```
Campalo, line 1335: Why is the `id` now optional for `InvoiceProductOutput `, it should be mandatory. The product id should always be in the BE response. I know it is not part of your pr, we may have missed it during @kolyose 's pr review.

lib/graphql/schema.ts
```diff
@@ -1355,13 +1332,44 @@ export type Invoice = {
...
+};
+
+export type DashboardInvoice = {
+  __typename?: 'DashboardInvoice';
+  id: Scalars['ID'];
+  status: InvoiceStatusType;
+  invoiceNumber?: Maybe<Scalars['Int']>;
+  dueDate?: Maybe<Scalars['DateTime']>;
+  paidAt?: Maybe<Scalars['DateTime']>;
+  amount?: Maybe<Scalars['Int']>;
```
Campalo, line 1362: Should it rather be `DECIMAL`? the total amount of the invoice may decimal most of the time, like shown on the mockup:
Screenshot 2021-04-13 at 14 52 21

amodesKontist: We always store amounts in cents

Campalo: Right, easier to make calculation in cents 👍

lib/graphql/schema.ts
amodesKontist: In BE its declared like this:
`@Arg("pageNumber") pageNumber: number`

If you search for 'float' in the repo, there are multiple places where its the same, so I think its fine 🤔

lib/graphql/schema.ts
amodesKontist: In `services/backendService/src/core/graphql/types/invoicing/product.ts` I see
```
export class UserProductInput {
  @Field({ nullable: true })
  id?: string;
  ...
  ```
  
  I think it was done in order to be able to use the same code for input and output. And for input id is optional

Comment (Campalo) [reply lib/graphql/schema.ts:1335]: I saw it too, but then we're creating two types based on the `InvoiceProduct` which itself extends `UserProductInput`
Those two types are exactly the same, so if we decide that the id is optional for both then we can just merge them into one.  But then I remembered that Sergey had types issues, and that's why we had to duplicate them, but I can find the conversation about it in ou slack channel 😅 
```
@InputType()
export class InvoiceProductInput extends InvoiceProduct {}

@ObjectType()
export class InvoiceProductOutput extends InvoiceProduct {}
```

Campalo: It can be fixed later, as discussed here: https://kontist.slack.com/archives/G01J0BJAA3U/p1618321820029700

lib/graphql/schema.ts
Campalo: Agree. I think it is fine too.

Campalo (approved):

amodesKontist:
I'll create new branch as I am too lazy for managing merge conflicts 😅

PR #191, opened 2021-04-14
Re-generate the schema

kolyose:
Updates the schema with new Invoicing product types


lera (approved):
I missed those types in my last PR 🙁

PR #218, opened 2021-06-18
Get declaration stats

lera:
returns amount to pay and transactions for requested declaration period, grouped by elster category

### DIFF
--- a/lib/graphql/declaration.ts
+++ b/lib/graphql/declaration.ts
@@ -6,6 +6,8 @@ import {
   AccountDeclarationsArgs,
   Declaration as DeclarationModel,
   MutationSubmitDeclarationArgs,
+  AccountDeclarationStatsArgs,
+  DeclarationStats,
 } from "./schema";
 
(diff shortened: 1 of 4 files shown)


lib/graphql/declaration.ts
```diff
@@ -50,6 +52,51 @@ const SUBMIT_DECLARATION = `
...
+          year: $year
+        ) {
+          amount
+          uncategorized {
+            id
+            amount
+            name
+            purpose
+            valutaDate
+            userSelectedBookingDate
```
temhota, line 70: Seems like this should be `selectedBookingDate` and  not `userSelectedBookingDate`
FE is complaining about it

lera: but it was already changed to selectedBookingDate🤔

lera: here https://github.com/kontist/js-sdk/pull/220/files

lera: ah! but not in the method itself 🤦‍♀️

PR #219, opened 2021-06-18
Extend Account with Available balance

lera:


lib/graphql/schema.flow.js
```diff
@@ -22,6 +22,7 @@ export type Account = {|
   publicId: $ElementType<Scalars, 'ID'>,
   iban: $ElementType<Scalars, 'String'>,
   cardHolderRepresentation?: ?$ElementType<Scalars, 'String'>,
+  availableBalance: $ElementType<Scalars, 'Float'>,
```
p-janik, line 25: it should be Int

lera: fixed

p-janik (approved):

PR #222, opened 2021-06-25
Remove deprecated mutation

lera:
`transaction.update` is used instead


amodesKontist:
Should we add a note to `CHANGELOG.md`, as its breaking change ?

lera:
yes, when we issue release

amodesKontist (approved):

PR #234, opened 2021-07-14
Add setTimestamp mutation

amodesKontist:
Adds helper for setting timestamp.
This BE PR is needed
https://github.com/kontist/backend/pull/10102


amodesKontist:
Will be most likely not needed

PR #241, opened 2021-08-13
Rename elsterCode to categoryCode

haipham-kontist:
Rename `elsterCode` to `categoryCode`, while keep `elsterCode` for backward compatibility.

### DIFF
--- a/lib/graphql/schema.flow.js
+++ b/lib/graphql/schema.flow.js
@@ -1929,6 +1929,7 @@ export type Transaction = {|
   foreignCurrency?: ?$ElementType<Scalars, 'String'>,
   originalAmount?: ?$ElementType<Scalars, 'Int'>,
   categoryCode?: ?$ElementType<Scalars, 'String'>,
+  elsterCode?: ?$ElementType<Scalars, 'String'>,
   categoryCodeTranslation?: ?$ElementType<Scalars, 'String'>,
   elsterCodeTranslation?: ?$ElementType<Scalars, 'String'>,
   recurlyInvoiceNumber?: ?$ElementType<Scalars, 'String'>,
--- a/lib/graphql/schema.flow.js
+++ b/lib/graphql/schema.flow.js
@@ -333,9 +333,20 @@ export type CategorizationType = $Values<typeof CategorizationTypeValues>;
 
 export type CategorizeTransactionForDeclarationResponse = {|
   __typename?: 'CategorizeTransactionForDeclarationResponse',
+  categoryCode?: ?$ElementType<Scalars, 'String'>,
   elsterCode?: ?$ElementType<Scalars, 'String'>,
   category?: ?$ElementType<Scalars, 'String'>,
   date?: ?$ElementType<Scalars, 'String'>,
+|};
+
+export type CategoryGroup = {|
+  __typename?: 'CategoryGroup',
+  categoryCode: $ElementType<Scalars, 'String'>,
+  elsterCode: $ElementType<Scalars, 'String'>,
+  amount: $ElementType<Scalars, 'Int'>,
+  categoryCodeTranslation: $ElementType<Scalars, 'String'>,
+  elsterCodeTranslation: $ElementType<Scalars, 'String'>,
+  transactions: Array<TransactionForAccountingView>,
 |};
 
 export type Client = {|
@@ -522,7 +533,8 @@ export type Declaration = {|
 export type DeclarationStats = {|
   __typename?: 'DeclarationStats',
   amount: $ElementType<Scalars, 'Int'>,
-  elsterGroups: Array<ElsterGroup>,
+  categoryGroups: Array<CategoryGroup>,
+  elsterGroups: Array<CategoryGroup>,
   uncategorized: Array<TransactionForAccountingView>,
 |};
 
@@ -563,14 +575,6 @@ export const DocumentTypeValues = Object.freeze({
 
 
 export type DocumentType = $Values<typeof DocumentTypeValues>;
-
-export type ElsterGroup = {|
-  __typename?: 'ElsterGroup',
-  elsterCode: $ElementType<Scalars, 'String'>,
-  amount: $ElementType<Scalars, 'Int'>,
-  elsterCodeTranslation: $ElementType<Scalars, 'String'>,
-  transactions: Array<TransactionForAccountingView>,
-|};
 
 export type FormDataPair = {|
   __typename?: 'FormDataPair',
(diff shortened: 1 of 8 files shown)


tests/auth/tokenManager.spec.ts
```diff
@@ -23,7 +23,7 @@ describe("Auth: TokenManager", () => {
       const codeChallenge = "xc3uY4-XMuobNWXzzfEqbYx3rUYBH69_zu4EFQIJH8w";
       const codeChallengeMethod = "S256";
       const url = await client.auth.tokenManager.getAuthUri();
-      const expectedUrl = `${Constants.KONTIST_API_BASE_URL}/api/oauth/authorize?client_id=${clientId}&redirect_uri=https%3A%2F%2Flocalhost%3A3000%2Fauth%2Fcallback&scope=transactions&response_type=code&state=25843739712322056&code_challenge=${codeChallenge}&code_challenge_method=${codeChallengeMethod}`;
+      const expectedUrl = `${Constants.KONTIST_API_BASE_URL}/api/oauth/authorize?client_id=${clientId}&redirect_uri=https%3A%2F%2Flocalhost%3A3000%2Fauth%2Fcallback&response_type=code&state=25843739712322056&scope=transactions&code_challenge=${codeChallenge}&code_challenge_method=${codeChallengeMethod}`;
```
haipham-kontist, line 26: I'm not sure why these tests are failing on my local. I just rearranged the query params to fix it.

lib/graphql/schema.flow.js
```diff
@@ -333,11 +333,22 @@ export type CategorizationType = $Values<typeof CategorizationTypeValues>;
 
 export type CategorizeTransactionForDeclarationResponse = {|
   __typename?: 'CategorizeTransactionForDeclarationResponse',
+  categoryCode?: ?$ElementType<Scalars, 'String'>,
   elsterCode?: ?$ElementType<Scalars, 'String'>,
```
haipham-kontist, line 337: The file is automatically generated. Because the backend supports backward compatibility, therefore we still have `elsterCode` in request/response.

lib/graphql/schema.flow.js
```diff
@@ -1257,6 +1261,7 @@ export type MutationUpsertProductsArgs = {|
 
 export type MutationCategorizeTransactionForDeclarationArgs = {|
   id: $ElementType<Scalars, 'ID'>,
+  categoryCode?: ?$ElementType<Scalars, 'String'>,
   elsterCode?: ?$ElementType<Scalars, 'String'>,
```
haipham-kontist, line 1265: The file is automatically generated. Because the backend supports backward compatibility, therefore we still have `elsterCode` in request/response.

lib/graphql/schema.flow.js
```diff
@@ -333,11 +333,22 @@ export type CategorizationType = $Values<typeof CategorizationTypeValues>;
...
   category?: ?$ElementType<Scalars, 'String'>,
   date?: ?$ElementType<Scalars, 'String'>,
 |};
 
+export type CategoryGroup = {|
+  __typename?: 'CategoryGroup',
+  categoryCode: $ElementType<Scalars, 'String'>,
+  elsterCode: $ElementType<Scalars, 'String'>,
```
fabribet, line 345: ❓ Do we need `elsterCode` in this type? I would have expected that when this is used, elster code is no longer needed

lib/graphql/schema.flow.js
```diff
@@ -564,14 +576,6 @@ export const DocumentTypeValues = Object.freeze({
 
 export type DocumentType = $Values<typeof DocumentTypeValues>;
 
-export type ElsterGroup = {|
```
fabribet, line 567: ❓ Should this be gone ? Are we not using it anywhere?

lib/graphql/schema.ts
```diff
@@ -530,14 +542,6 @@ export enum DocumentType {
   Invoice = 'INVOICE'
 }
 
-export type ElsterGroup = {
```
fabribet, line 533: ❓ Same question as before

fabribet (commented):
I left some questions for my understanding.

lib/graphql/schema.flow.js
haipham-kontist: all of this is automatically generated from backend. Because backend has `elsterCode` in request/response, so it's generated like this.

lib/graphql/schema.flow.js
haipham-kontist: all of this is automatically generated from backend. Because backend has `elsterCode` in request/response, so it's generated like this.

lib/graphql/schema.ts
haipham-kontist: all of this is automatically generated from backend. Because backend has `elsterCode` in request/response, so it's generated like this.

lib/graphql/schema.flow.js
fabribet: I know this is auto-generated, but my questions is more about what we are setting on the back-end. I am unsure why we have elsterCode inside CategoryGroup. Maybe I got lost.

But the concept behind my question applies to all comments

haipham-kontist: I see. It is like this because in backend, `categoryGroups` and `elsterGroups` share the same type.

```
@Field((type) => [CategoryGroup])
categoryGroups: CategoryGroup[];

// @deprecated - `elsterCode` should no longer be used. Use `categoryCode` instead.
@Field((type) => [CategoryGroup])
elsterGroups: CategoryGroup[];
```

then

```
class CategoryGroup {
  @Field((type) => String)
  categoryCode: string;

  // @deprecated - `elsterCode` should no longer be used. Use `categoryCode` instead.
  @Field((type) => String)
  elsterCode: string;

  @Field((type) => String)
  categoryCodeTranslation: string;

  // @deprecated - `elsterCode` should no longer be used. Use `categoryCode` instead.
  @Field((type) => String)
  elsterCodeTranslation: string;
}
```

haipham-kontist: If you wonder why they share the same type, then my answer would be to make it safe for backward compatibility and easy to clean up in the future.

haipham-kontist: But maybe I'm too cautious. Let me try to make it cleaner.

haipham-kontist: Ah, I see why I did this. It's not just the type, it's the whole calculation behind it, namely `getDeclarationTransactions`.

fabribet: Thanks for taking the time checking

lib/graphql/schema.flow.js
fabribet: So, just to double check. Are we not using ElsterGroup anymore? 🤔

haipham-kontist: no we don't, neither in `webapp` nor `app`

fabribet (approved):

p-janik (approved):

PR #246, opened 2021-09-07
Extend transaction response

lera:


p-janik:
I think we can update default test payload in tests too

p-janik (approved):

PR #252, opened 2021-10-26
Add document model

p-janik:


.nycrc
```diff
@@ -1,7 +1,7 @@
 {
     "cache": false,
     "check-coverage": true,
-    "branches": 93,
+    "branches": 92,
```
amodesKontist, line 4: what is this change doing?

lib/graphql/document.ts
```diff
@@ -0,0 +1,68 @@
...
+url
+note
+createdAt
+`;
+
+const FETCH_DOCUMENTS = `
+  query {
+    viewer {
+      documents {
+        ${DOCUMENT_FIELDS}
```
amodesKontist, line 21: I wonder sometimes, if we hardcode the response values like this in sdk, isn't it destroying a bit a benefit of gql:
Having flexibility and beeing able to decide on the client what particular fields we want to have in the response.

.nycrc
p-janik: it's making `npm run cover` pass. https://github.com/kontist/js-sdk/pull/252/commits/6a5e467404a18d4e77bdb8bf4c33491a3ee00d97

lib/graphql/document.ts
p-janik: Updated with https://github.com/kontist/js-sdk/pull/252/commits/8a7b7559e645b0339901ffb45cf4c4c34587c2d9

amodesKontist (approved):

PR #262, opened 2022-01-31
Update deps

johannespfeiffer:
Follow-up PR to update dependencies.

If merged, I suggest that we generate fresh schema files, as they differ quite a lot.


lera (approved):

lib/utils/index.ts
```diff
@@ -1,2 +1,7 @@
-export * from "./authorizeSilently";
+import { authorizeSilently } from "./authorizeSilently";
+
 export * from "./serializeGraphQLError";
+
+export const utils = {
+    authorizeSilently,
+}
```
sinahashemy, line 7: 🛠 isn't need for one newline here?

package.json
```diff
@@ -15,39 +15,39 @@
...
+    "tslint": "^6.1.3",
+    "typescript": "^4.5.5",
+    "webpack": "^4.46.0",
+    "webpack-cli": "^4.9.2"
   },
   "bin": {
     "kontist": "./cli/index.js"
   }
-}
+}
```
sinahashemy, line 53: 🛠 isn't need for one newline here?

tests/graphql/account.spec.ts
```diff
@@ -1,9 +1,9 @@
-import { expect } from "chai";
 import * as sinon from "sinon";
 
+import { Account } from "../../lib/graphql/account";
 import { Client } from "../../lib";
 import { KontistSDKError } from "../../lib/errors";
-import { Account } from "../../lib/graphql/account";
+import { expect } from "chai";
```
sinahashemy, line 6: 🛠 in other repositories we kept the external library at top of the file, then the internal libraries!

tests/graphql/user.spec.ts
```diff
@@ -1,9 +1,9 @@
-import { expect } from "chai";
 import * as sinon from "sinon";
 
 import { Client } from "../../lib";
 import { KontistSDKError } from "../../lib/errors";
 import { User } from "../../lib/graphql/user";
+import { expect } from "chai";
```
sinahashemy, line 6: 💭 like the previous comment.

sinahashemy (approved):
Thank you for updating the libraries 👍. I leave some comments.

johannespfeiffer:
@sinahashemy valid point with the style changes. We do use `tslint` in this project and just mostly stick to the latest defaults via `tslint:latest`. Probably that is where it came from.

PR #263, opened 2022-02-02
Update schema

sinahashemy:
This PR has been generated for updating schema due to new updating in the deps that create the schema by the [Update deps](https://github.com/kontist/js-sdk/pull/262)


johannespfeiffer:
This changes are expected, because of the recent @graphql-codegen. As discussed, please make sure to test this new version inside of the backend. One way to test this locally would be to use `npm link`.

johannespfeiffer (approved):

PR #268, opened 2022-03-15
Update schema

amodes:


lera:
why do you have so many changes? could you include only yours?

lera (approved):

PR #273, opened 2022-03-24
Add TaxNumber and BusinessAddress to Client Model

sinahashemy:
Extend client models to add `TaxNumber` and `BusinessAddress`.


dmytroyarmak (approved):
👍

PR #281, opened 2022-05-04
Add emailDocument matching helpers

amodes:
Adds emailDocument related logic:

`fetchAll`
`fetchOne`
`matchEmailDocumentToTransaction`

### DIFF
--- a/lib/graphql/emailDocument.ts
+++ b/lib/graphql/emailDocument.ts
@@ -1,6 +1,8 @@
 import { GraphQLClient } from "./client";
 import {
   EmailDocument as EmailDocumentModel,
+  MutationMatchEmailDocumentToTransactionArgs,
+  MutationResult,
   Query,
   Transaction as TransactionModel,
   UserEmailDocumentArgs,
@@ -28,7 +30,7 @@ const DEFAULT_EMAIL_DOCUMENT_MATCH_FIELDS = [
   "description",
   "valutaDate",
   "personalNote",
-  "type"
+  "type",
 ] as TransactionProps[];
 
 const DEFAULT_EMAIL_DOCUMENT_FIELDS = ["id"] as EmailDocumentProps[];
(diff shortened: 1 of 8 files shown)


lib/graphql/emailDocument.ts
```diff
@@ -0,0 +1,115 @@
...
+  Query,
+  Transaction as TransactionModel,
+  UserEmailDocumentArgs,
+  UserEmailDocumentsArgs,
+} from "./schema";
+
+type EmailDocumentProps = keyof EmailDocumentModel;
+type TransactionProps = keyof TransactionModel;
+
+const DEFAULT_EMAIL_DOCUMENTS_FIELDS = [
```
amodes, line 15: I added default fields which I think are needed for our webapp usecase.
If we decide later on that we need other fields we can still do it by using optional `fields` and `matchFields`

p-janik (approved):
nit: maybe we could limit changes applied in schema files to the ones related to this PR

Regarding lowering branch coverage, I think it's fine. Otherwise we would need to add tests for `?.` cases

johannespfeiffer:
I vote against lowering the coverage.
We started with 100% - but then we keep lowering it 1 by 1.

Of course - 100% does not mean we cover every case. But the other way around is helpful: If we have a coverage less than 100%, we can be sure that we did NOT cover every case.

amodes:
> nit: maybe we could limit changes applied in schema files to the ones related to this PR
> 
> Regarding lowering branch coverage, I think it's fine. Otherwise we would need to add tests for `?.` cases

Those updates in the schema where actually all related because there is an issue:
schema:generate generates `transactionMatches` which is a field which is just available for the kontax resolver.

I created a housekeeping ticket for it https://kontist.atlassian.net/jira/software/c/projects/HOUS/boards/39?modal=detail&selectedIssue=HOUS-154

PR #300, opened 2022-07-06
Generate schema, update GET_USER query for TSB-11

UgurGumushan:
https://kontist.atlassian.net/browse/TSB-11
The structure user will change after changes on the backend after a followup pr


Campalo (approved):
Before releasing it, please make sure that the change of name from ´screeningStatus´ to  ´screeningProgress´ is not breaking the signup flow on the webapp and the app. It might make sense to keep both the old and the new name to start with.

UgurGumushan:
> Before releasing it, please make sure that the change of name from ´screeningStatus´ to ´screeningProgress´ is not breaking the signup flow on the webapp and the app. It might make sense to keep both the old and the new name to start with.

Yes, we'll need to keep both. I'll be back with some changes.

4lphaquanta (approved):

PR #308, opened 2022-08-03
Transaction filter presets

p-janik:
New search presets for fetching tax/vat related transactons with missing receipts

Backend PR: https://github.com/kontist/backend/pull/12205

### DIFF
--- a/lib/graphql/transaction.ts
+++ b/lib/graphql/transaction.ts
@@ -522,7 +522,7 @@ export class Transaction extends IterableModel<TransactionModel> {
     return result.viewer?.mainAccount?.transactionsCSV ?? "";
   }
 
-  public async fetchFilterPresets(): Promise<FilterPreset[]> {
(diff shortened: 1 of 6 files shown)


lib/graphql/transaction.ts
```diff
@@ -501,4 +521,9 @@ export class Transaction extends IterableModel<TransactionModel> {
     const result: Query = await this.client.rawQuery(FETCH_TRANSACTIONS_CSV, args);
     return result.viewer?.mainAccount?.transactionsCSV ?? "";
   }
+
+  public async fetchFilterPresets(): Promise<FilterPreset[]> {
```
dddenis, line 525: let's put all possible `FilterPreset` implementations as a union here, so that the client can differentiate them by the `__typename` field

```suggestion
  public async fetchFilterPresets(): Promise<(GenericFilterPreset | MissingTaxAssetsFilterPreset)[]> {
```

dddenis (approved):

PR #309, opened 2022-09-08
Wann 531/updates schema for seizure protection

DanaNussair:


lib/graphql/account.ts
```diff
@@ -37,6 +37,44 @@ const GET_ACCOUNT_STATS = `query {
   }
 }`;
 
+const GET_SOLARIS_BALANCE = `query {
+  viewer {
+    mainAccount {
+      solarisBalance {
+          balance {
+            value,
```
mkhstar, line 45: The commas in the query can be removed to make everything consistent, even though graphql works with commas but it is not recommended

mkhstar (commented):
Looks good to me, but i need to test it after my backend builds 👍🏾

mkhstar (approved):
Lgtm

PR #327, opened 2023-01-26
Add methods for encrypted PIN change

dmytroyarmak:
Add `getPinKey` and `changePINEncrypted` to cards model to be used during encrypted PIN change flow


p-janik (approved):

Campalo (commented):
Should we mark both`ChangePIN()` and `ConfirmChangePIN` mutations deprecated? and then remove them in a follow up pr, once everything is released?
https://github.com/kontist/js-sdk/pull/327#diff-b1af1caa11d12d6efb383ae3a365ab2abfbc5d688aef224c2a5ad67cfad1697eR405

Campalo (commented):
At the end of last year, we discussed that we should stop updating the sdk and rather use graphql queries. Should we then do it here too for those two new methods (even though you already did the work)? 
--> https://agerascom.slack.com/archives/G049Q6A2YHJ/p1675075543393339?thread_ts=1675072925.767249&cid=G049Q6A2YHJ : _"At least when we create new queries, I suggest to skip the sdk and instead fetch only the required fields via a raw graphql query in the apps. - Joe said"._

sinahashemy (approved):

dmytroyarmak:
> Should we mark both`ChangePIN()` and `ConfirmChangePIN()` mutations deprecated as they are not used anymore in the mobile app? and create a subtask to remove them in a follow up pr, once everything is released? [#327 (comment)](https://github.com/kontist/js-sdk/pull/327#diff-b1af1caa11d12d6efb383ae3a365ab2abfbc5d688aef224c2a5ad67cfad1697eR405)

Yes, we would need to cleanup and deprecate old pin change after we migrated to the new one

> At the end of last year, we discussed that we should stop updating the sdk and rather use graphql queries. Should we then do it here too for those two new methods (even though you already did the work)? --> https://agerascom.slack.com/archives/G049Q6A2YHJ/p1675075543393339?thread_ts=1675072925.767249&cid=G049Q6A2YHJ : _"At least when we create new queries, I suggest to skip the sdk and instead fetch only the required fields via a raw graphql query in the apps. - Joe said"._

I think it's still make sense to have it in the sdk because it's a replacement for already existing functions

PR #334, opened 2023-02-20
Card limits

lera:


lib/graphql/schema.ts
```diff
@@ -295,8 +295,8 @@ export type CardLimit = {
 };
 
 export type CardLimitInput = {
-  maxAmountCents: Scalars['Float'];
-  maxTransactions: Scalars['Float'];
+  maxAmountCents?: InputMaybe<Scalars['Float']>;
+  maxTransactions?: InputMaybe<Scalars['Float']>;
```
UgurGumushan, line 299: Maybe it shouldn't be float.
Not a blocker if it works.

UgurGumushan (approved):

lib/graphql/schema.ts
lera: 🤔 it's automatically generated type and on BE it's number

PR #347, opened 2023-05-02
Update schema

caglaralkiss:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kontist",
-  "version": "0.49.10",
```
caglaralkiss, line 3: I increased patch version by 2 in here because there is an inconsistency between version at package.json and [npm-registry](https://www.npmjs.com/package/kontist?activeTab=readme). An accidental [push tag](https://github.com/kontist/js-sdk/releases/tag/v0.49.11) caused this inconsistency but it is [problematic](https://stackoverflow.com/questions/15056377/how-do-i-update-an-npm-module-that-i-published) to override package with specific version so I ended up with increasing patch version number by 2.

sinahashemy: OK good, do you want to bump the version with this PR? 
Usually, we first merge the changes and then bump the version with a separate PR.

dmytroyarmak (approved):

package.json
caglaralkiss: I didn't know this. I will revert version changes then open a new PR for version update then. Thank you 👍

dmytroyarmak: @caglaralkiss It's not critical since this repo is not very active, so I think there is no need to revert

dmytroyarmak: it's just better to bump version in separate PR to avoid conflicts when several PRs are bumping versions at the same time and it's actually what happened with your PR and https://github.com/kontist/js-sdk/pull/349 🙂

PR #358, opened 2023-05-24
Bump version to 0.49.17

sinahashemy:
Bump version to 0.49.17


p-janik (commented):
Did you update it via `npm version patch`? I would expect also a change in lock file and tag to be created (`git push --folow-tags`)

p-janik (approved):

PR #365, opened 2023-06-07
generate the required type for user business assets

EslaM-M:


OgulcanCelik (commented):
thank you for the light speed PR, there are some unrelated comment deletions would be great if you could re-add them 👍 also you need to update the package version in `package.json` 😄

EslaM-M:
> thank you for the light speed PR, there are some unrelated comment deletions would be great if you could re-add them 👍 also you need to update the package version in `package.json` 😄

@OgulcanCelik done 🙏

OgulcanCelik (approved):
🚀

sinahashemy (approved):
👍, would be better to bump the version in a separated pr as a release pr, it will prevent conflicts on the multiple pr simultaneously.

PR #371, opened 2023-06-14
COM-1363 External transaction categorization by bk users

caglaralkiss:


caglaralkiss:
It seems `yarn.lock` file is [accidentally added](https://github.com/kontist/js-sdk/pull/365/files) where we manage this package with npm so I removed it.

dmytroyarmak (approved):

PR #399, opened 2023-08-15
Revert "Require node 16"

p-janik:
Reverts kontist/js-sdk#397

When trying to install on app with node 18, err is thrown.


UgurGumushan (approved):

UgurGumushan:
On revert, it goes to 12 from 16 and not 18
You may not need this change

PR #408, opened 2023-09-11
Dev-78 Remove planCopy from subscriptionPlans

Campalo:
🧹 
Now that we're using the webview for the mobile signup, we can use the plan copy coming from Lokalise for both webapp and mobile ([see FE pr for this)](https://github.com/kontist/webapp/pull/4408). 
So this pr removes the planCopy from subscriptionPlans query

### DIFF
--- a/lib/graphql/schema.ts
+++ b/lib/graphql/schema.ts
@@ -1043,6 +1043,11 @@ export enum GrantType {
   RefreshToken = 'REFRESH_TOKEN'
 }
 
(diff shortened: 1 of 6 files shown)


sinahashemy (approved):

dmytroyarmak (approved):

lib/graphql/schema.ts
Campalo, line 1: Changes form the schema are not related to this pr

PR #411, opened 2023-09-20
Dev 78/deprecate old copy fields for backward compatibility

Campalo:
### Description

This pr put backs the old fields in the `fetchPlans`query and update the schema to take into account the deprecated old copy fields which are not used anymore on the webapp (copy is coming from Lokalise now) ; we deprecate them without removing them for backward compatibility on the mobile app because some of our users are using an app version which uses a former version of our SDK (where those fields are still queried which results in errors in our BE logs, but does not impact the user experience).

At some point the users will be forced to update to a new app version using at least this current SDK version ; from then on will be able to remove those deprecated fields in both SDK and BE. 

### Type of change

- [x] Chore (technical improvement, i.e. cleanup, refactor, etc.)

### DIFF
--- a/lib/graphql/schema.flow.js
+++ b/lib/graphql/schema.flow.js
@@ -1435,6 +1435,7 @@ export type Mutation = {|
   createReview: CreateReviewResponse,
   /** Create user's taxNumber */
   createTaxNumber: TaxNumber,
+  createTopUp: TopUpCreationResult,
   /** Create a transaction Asset and obtain an upload config */
   createTransactionAsset: CreateAssetResponse,
   /** Create transaction splits */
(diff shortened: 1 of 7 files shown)


lib/graphql/schema.flow.js
```diff
@@ -3100,13 +3120,18 @@ export type SubscriptionFeatureGroup = {|
...
+  description?: ?$ElementType<Scalars, 'String'>,
+  /** @deprecated For backwards compatibility on mobile only. From now on use the features copy coming from Lokalise instead. */
+  featureGroups?: ?Array<SubscriptionFeatureGroup>,
+  /** @deprecated For backwards compatibility on mobile only. */
   featuresToggleLabel?: ?$ElementType<Scalars, 'String'>,
   fee: Money,
   subtitle?: ?$ElementType<Scalars, 'String'>,
-  title: $ElementType<Scalars, 'String'>,
+  /** @deprecated For backwards compatibility on mobile only. From now on use the title copy coming from Lokalise instead. */
+  title?: ?$ElementType<Scalars, 'String'>,
```
Campalo, line 3134: Old copy fields are now optional and deprecated

dmytroyarmak (approved):

PR #421, opened 2023-10-31
change schema for biztax plan

OgulcanCelik:
changes for renaming biztax -> complete

### DIFF
--- a/lib/graphql/schema.flow.js
+++ b/lib/graphql/schema.flow.js
@@ -304,12 +304,6 @@ export const BatchTransferStatusValues = Object.freeze({
 
 export type BatchTransferStatus = $Values<typeof BatchTransferStatusValues>;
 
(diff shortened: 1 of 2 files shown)


Campalo (commented):
⚠️  We should be careful with the renaming, especially for the mobile app. Mobile app users do not always use the latest app version and so do not use the latest sdk version. So we will need to force the mobile user to update to the latest SDK version that includes the plan renaming, else the mobile app will break.

OgulcanCelik:
> ⚠️  We should be careful with the renaming, especially for the mobile app. Mobile app users do not always use the latest app version and so do not use the latest sdk version. So we will need to force the mobile user to update to the latest SDK version that includes the plan renaming, else the mobile app will break.

BizTax isn't released yet so it should be fine.

PR #422, opened 2023-11-24
COM-179 Update schema to extend transaction filters

sinahashemy:
Update schema to extend transaction filters

### DIFF
--- a/lib/graphql/schema.flow.js
+++ b/lib/graphql/schema.flow.js
@@ -310,6 +310,40 @@ export type BizTaxBookkeepingConfirmation = {|
   year: $ElementType<Scalars, 'Int'>,
 |};
 
+export type BizTaxCarUsageEuerRows = {|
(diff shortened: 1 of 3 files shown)


Campalo (approved):
💭  I remember we said that we should not update our sdk with new features anymore. On the frontend side, do we then plan to define the types again for `BizTaxCarUsageEuerRows` and the other new types rather than use their Schema definition `Schema. BizTaxCarUsageEuerRows`?

sinahashemy:
> 💭 I remember we said that we should not update our sdk with new features anymore. On the frontend side, do we then plan to define the types again for `BizTaxCarUsageEuerRows` and the other new types rather than use their Schema definition `Schema. BizTaxCarUsageEuerRows`?

Yes, you are right, for the TransactionFilter we should update the sdk, other updates got received automatically, we don't use it.

PR #443, opened 2024-03-01
DEV-269 Use `isomorphic-ws` instead of `ws`

dmytroyarmak:
Use `isomorphic-ws` instead of `ws` to fix problem with `webpack` v5 complaining about native module usage by `ws` (it's intender to be run only on Node.js)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,9 +23,10 @@
     "graphql": "^16.3.0",
     "graphql-request": "^4.0.0",
     "graphql-ws": "^5.5.5",
+    "isomorphic-ws": "^5.0.0",
     "js-sha256": "^0.9.0",
     "querystring-es3": "^0.2.1",
-    "ws": "^8.4.2"
+    "ws": "^8.16.0"
   },
   "overrides": {
     "tough-cookie": "4.1.3"
--- a/tests/graphql/client.spec.ts
+++ b/tests/graphql/client.spec.ts
@@ -1,5 +1,4 @@
 import * as sinon from "sinon";
-import * as ws from "ws";
 
 import { GraphQLError, UserUnauthorizedError } from "../../lib/errors";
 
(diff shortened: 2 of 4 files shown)


tests/graphql/client.spec.ts
```diff
@@ -552,27 +551,6 @@ describe("createSubscriptionClient", () => {
...
-
-      it("should use the native browser WebSocket implementation", () => {
-        createClientStub.resetHistory();
-
-        client.graphQL.createSubscriptionClient();
-
-        expect(createClientStub.callCount).to.equal(1);
-        expect(createClientStub.getCall(0).args[0].webSocketImpl.fake).to.equal(true);
-      });
-    });
```
OgulcanCelik, line 575: 🤔 why we are removing test? we can still verify if `isomorphic-ws` using native WebSocket with this

dmytroyarmak: This test case was specifically for `typeof window === "undefined"` condition that was in our code base, but now it's removed and it's now internal logic of `isomorphic-ws`. They actually handle it in more graceful way by having `browser` entry point in `package.json` that loads very tiny JS wrapper to normalize native websocket and for node.js it loads completely different file.

Anyway, I've extendede previous test case to verify that we pass `WebSocket` from `isomorphic-ws` to `createClient`: d5600cc

sinahashemy (approved):

package.json
```diff
@@ -23,9 +23,10 @@
     "graphql": "^16.3.0",
     "graphql-request": "^4.0.0",
     "graphql-ws": "^5.5.5",
+    "isomorphic-ws": "^5.0.0",
     "js-sha256": "^0.9.0",
     "querystring-es3": "^0.2.1",
-    "ws": "^8.4.2"
+    "ws": "^8.16.0"
```
Campalo, line 29: ❔ Why did you downgrade `ws` library? 
Also could it be totally replaced by `isomorphic-ws` library?

Campalo (approved):

package.json
dmytroyarmak: `ws` is actually upgraded from 4 to 16 😉 It's just unusual to read when version number has 2 digits

`isomorphic-ws` internally uses `ws` for NodeJS and they require to have `ws` package installed as "peer dependency".

Campalo: > ws is actually upgraded from 4 to 16 😉 It's just unusual to read when version number has 2 digits
🙈

PR #454, opened 2024-07-29
upgrade vulnerable packages, upgrade node because it is necessary

UgurGumushan:


UgurGumushan:
@lera we can generate schema after these changes but I have not checked if it's still usable on the other projects. checking...

lera (approved):

